Issue #1, opened 2022-02-24
Suggestion: simulating species tutorial narrative


jeromekelleher:
I think a good way of structuring this paper would be to couch it as a "So, you want to simulate your favourite species, eh?  Here's what you need" narrative. The discussion about how we do things in stdpopsim and the lessons we've learned through the hackathon can be incorporated into this narrative, without making the whole thing *about* stdpopsim and how we maintain its catalog. We can have a section in there where we discuss the current contents of the catalog and the processes that we have for maintaining it. I think this would generally be quite a useful paper, as it seems to me that it's generally not well understood what the components of a realistic simulation model are. 

Apologies if this is going over old ground - we can close this issue if there's already strong ideas on the narrative and I missed the memo!

lauterbur:
I like this idea a lot - it targets more toward utility to (new) users than being about stdpopsim itself. That said, I think it would also require more attention to the "but what if my favorite species doesn't have a chromosome-level assembly?" issue we've been wrangling with, that I've currently addressed with hand-waving and references to stdpopsim being "community-driven, hint hint." That could bring us back around to your yaml idea (https://github.com/popsim-consortium/stdpopsim/pull/1020) but I doubt the implementation timeline for that fits the preprint by ProbGen timeline. Thus the answer might be a disappointing "tough luck" or a somewhat less disappointing but still unsatisfying "stay tuned." 

What do you think?

jeromekelleher:
I think it's fine to say that doing simulations of species that don't have well characterised chromosomes (== good assemblies) is problematic. You just have to make some stuff up that seems plausible to you, ultimately :smile:

Put it another way: I don't think we need to apologise for the fact that not every species has a high quality assembly. We provide some help and automation for those that do (stdpopsim) but here's some general guidance on what you can do if that's not possible (do you really *need* to simulate a whole chromosome? does it actually matter for the question you're asking, or can just some unspecified length of genome do the job just as well?)

lauterbur:
Yes, I see how that could work, and might not actually take much restructuring from what we have now. I can give that a go later today or tomorrow unless there are other opinions.

@gregorgorjanc @izabelcavassim since you've already weighed in on Slack, thoughts?

gregorgorjanc:
Sounds good to me @lauterbur ! I think you have a very good start, but taking @jeromekelleher take into account is sensible

petrelharp:
[...]

PR #5, opened 2022-03-07
Required features to add a species to the catalog

izabelcavassim:
This is a rough draft for the session I chose to write about. I know a lot of it was discussed in our first issue #1  . I don't know how much of the "too many contigs" problem needs to be included in this paragraph (if not here we could do it in discussion). I also suggest including in discussion that the "too many contigs" problem won't be a problem in the near future with the advent of these many consortiums that are actively and massively producing better genome assemblies across vertebrates (at least)...


adding_species_manuscript.md
```diff
@@ -74,7 +74,7 @@ preamble: |
...
-- (2) It also requires that the species has a plausible and citeable **mutation rate** estimate. Both phylogenetic mutation rate [ref], germline *de novo* mutations, or mutation rates estimates based on mutation accumulation studies [ref] are welcomed. [commment: using the mutation rate of a closely related species is not a big deal, either??? I know you guys use the drosophila one for the anopheles].
+- (2) It is also required that the species has a plausible and citeable **mutation rate** estimate. Both phylogenetic mutation rate [ref], germline *de novo* mutations, or mutation rates estimates based on mutation accumulation studies [ref] are welcomed. [commment: using the mutation rate of a closely related species is not a big deal, either??? I know you guys use the drosophila one for the anopheles].
 - (4) Since both mutation and recombination shape the genetic diversity of genomes, stdpopsim also requests a **recombination rate** estimate for the species. Ideally, a chromosome level recombination map--e.g., allowing to more precisely infer the effect of selective interference--or at least a citeable single recombination rate estimate.
```
lauterbur: (3) and (4) swapped?

lauterbur:
```suggestion
- (4) Since both mutation and recombination shape the genetic diversity of genomes, stdpopsim also requires a **recombination rate** estimate for the species. Ideally, this should be a chromosome level recombination map, which allows more precise inference of the effect of selective interference. At minimum, it is a citeable single recombination rate estimate for the whole genome.
```

adding_species_manuscript.md
```diff
@@ -74,7 +74,7 @@ preamble: |
...
 - (3) A generation time estimate (in years).
```
lauterbur: Worth mentioning that this could be mutation rate * generation time?

lauterbur (changes requested):
Thanks, Izabel!

adding_species_manuscript.md
izabelcavassim: can it be?

lauterbur: No, I misread. I blame sleepiness.

PR #6, opened 2022-03-08
reflow outline re:issue #1

lauterbur:


adding_species_manuscript.md
```diff
@@ -27,102 +27,74 @@ preamble: |
...
+        -   we present the current method for adding species and demographic scenarios to the catalog 
+        -   and clarify the required genomic resources and QC process
+
+3.   Making a popgen simulation
+    - necessary resources:
+        - genome assembly
+            - this needs to be chromosome level, or nearly so
+            - contigs are not sufficient because:
+                - lack of linkage information limits their use for inference
+                - current limitations to the Python API/command line (eg. it complicates displaying and choosing what chromosome (contig) to use)
```
jeromekelleher: I don't think we need to dwell too much on this - we should focus on what it would really *mean* to do simulations of a bazillion contigs (not a lot, IMO) vs giving technical reasons for why it's not done in stdpopsim. I think @petrelharp had some very good text on that point that we should try to get in here.

jeromekelleher (commented):
Looks great @lauterbur, I like it a lot. My only feedback is to step back a bit from the technical details of why we don't want to add species with fragmented assemblies into stdpopsim, vs thinking about what it would mean to do those simulations and providing constructive advice for people working with such species.

adding_species_manuscript.md
lauterbur: Good advice (and I'd actually meant to delete those last points with specific reference to stdpopsim, whoops). @izabelcavassim has made some progress on the first parts of this section at https://github.com/popsim-consortium/adding-species-manuscript/pull/5 and I think I will ask @petrelharp to incorporate some of his ideas for the "what about species without good assemblies" section.

izabelcavassim: maybe we could have a session: "what if one of the features above required are missing":
like:

- if a good mutation rate estimation is missing ------> borrow it from the closest species one can find
- same as above for recombination rate (?)
- if a good assembly does not exist --------> select the most contiguous/largest contigs and
- if a demographic model does not exist ----> at least Ne
- if Ne does not exist ----> the closest related species?

At the end of the day, we want all those parameters to be explicitly documented/citeable such that if at some point they are updated in the literature (e.g., a study estimated a new mutation rate for the species in the catalog) they can also be updated in the catalog. I think this is the great advantage of this open-source community, and I think we could emphasize it as well.

lauterbur: Maybe we could have a table for this information? I would argue that how useful the closest related species is depends on its closeness and the feature. If the feature is mutation rate, well, people have been happy using 10^-8 for pretty much any mammal for years, so maybe it's not so bad to pick something reasonable and cross your fingers. Similar for recombination rate. Ne though can vary a huge amount between closely related species, and there are a lot of people who may want to do simulations on species of conservation concern with population sizes that are even more likely to differ from related species.

I think we should emphasize throughout that these features should be citeable and chosen with deliberation, even for the more general parts that aren't directly referencing stdpopsim. And it's a good point that we should also emphasize that anything that ends up in the catalog has the potential to be updated (because science).

izabelcavassim: awesome! Agreed!

lauterbur:
I added some things in response to the notes from this morning's meeting, as well as today's issues. Thoughts? Good to go so other folks can claim their sections, or needs a little more adjustment?

izabelcavassim:
I think that is well enough structured for my taste! Thanks!

adding_species_manuscript.md
```diff
@@ -25,104 +25,87 @@ preamble: |
...
-
+Outline
+
+# Introduction
+1.  Simulation is important for population genomics, both methods development and inference
+    - there is a building avalanche of genomic data for a variety of species, and a complementary flourishing of methods for population genomic inference
+        - so the number and variety of species requiring detailed and accurate simulation models is increasing quickly
+    -   coding population genetic simulation models can be arduous and error-prone, but important
+        -   provide training data for machine learning, ABC
+        -   for both methods development and empirical research
```
petrelharp, line 36:
```suggestion
        -   for both methods development and empirical research
        - "accurate" includes realistic demographic models and genetic maps as both can strongly impact the patterns of variation and haplotype/linkage
```

petrelharp: I think we want to make super clear what's important about whole-genome sims (and when it's not as important) - we'll have some discussion below, but we don't want to just say "more realism is better"/

adding_species_manuscript.md
```diff
@@ -25,104 +25,87 @@ preamble: |
...
+    - stdpopsim is a tool recently developed to provide easy access to simulation frameworks
+        -   avoid re-making common models
+        -   provide standard benchmarks for methods development and testing (see Adrion et al.)
+        -   improves reproducibility
```
petrelharp, line 41:
```suggestion
        -   improves reproducibility
        - lower the barrier to technical coding skills
```

adding_species_manuscript.md
```diff
@@ -136,67 +119,81 @@ Table of species/demographic scenarios added/worked on by community since origin
...
+    
+# Conclusion
+7.  Take-aways
+    - we present the basics for 
+        - 1) determining if a species-specific population genomic simulation is appropriate for the species/question
+        - 2) getting the required data
+        - 3) creating a population genomic simulation of that species in your simulator of choice
+        - 4) adding that species to the stdpopsim catalog
+        - we *do not* present tutorials for implementation in any particular simulator (there are lots of those)
+    - there is a definite need for more and more species-specific population genomic simulations
```
petrelharp, line 151:
```suggestion
    - For many species we need better genomic resources: assemblies and annotations would be community resources
    - there is a definite need for more and more species-specific population genomic simulations
```

adding_species_manuscript.md
lauterbur: Absolutely. I think it's a good idea to highlight it in the intro, with more detail later. Is this something you could pull into section 3 "Making a popgen simulation" after @izabelcavassim's part on the necessary resources?

adding_species_manuscript.md
lauterbur: Added table. Do you want to take point on filling it out?

PR #18, opened 2022-03-15
Missing parameters paragraph

izabelcavassim:
resolving issue #11 , might be helpful to get some input from @petrelharp.

Thanks

PR #19, opened 2022-03-16
drafts of various bits

petrelharp:
In somewhat bad git style, I've done *four* distinct things here. Apologies! But, I've kept them in separate commits (so, could break them out into separate PRs if we want), and I wrote them all at once, so didn't want to go through the trouble of making separate PRs that would then have to be stacked on top of each other.
The four things are:

1. tidying up the spacing in the document
2. writing a bit about "when do we need whole-genome simulations"
3. writing a bit about "what if we don't have chromosome-level assemblies"
4. turning the "list of things we need" into sort of a narrative.

Everything is very rough-draft-ey, but I'd propose to deal with big objections, if any, then merge this in and improve it. (That probably means rearranging bits, rewriting, etc).

### DIFF
--- a/adding_species_manuscript.md
+++ b/adding_species_manuscript.md
@@ -95,6 +95,8 @@ Outline
 
 # Tutorial
 
+TODO: general introduction to this section
+
 ##  Whole-genome simulations: when do we need them?
 
 The stdpopsim framework aims to make it easy to simulate *whole chromosomes*,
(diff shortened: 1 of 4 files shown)


jeromekelleher (approved):
I had a read through, looks great!

petrelharp:
Ok - I've merged in #18, @izabelcavassim - shall we merge this and close that one?

izabelcavassim:
yes @petrelharp, fine by me. Let me know if you need any other input from me. Happy to write.

adding_species_manuscript.md
```diff
@@ -162,28 +164,113 @@ with stdpopsim.
...
-
-- (6) **Genome annotation**, such as a GFF3/GFF format [ref] containing information about the coordinates of coding and noncoding regions; and the position of specific genes. 
-
-- (7) A citeable **distribution of fitness effects** (DFE, REF) for the species. This feature is particularly useful when simulating genomes under non-neutral models. 
-
-    - draw from https://github.com/popsim-consortium/workshops/blob/main/adding_species/contributing.ipynb without specifics for stdpopsim 
-    - emphasize throughout that these features should be citeable and chosen with deliberation **PETER**
-
-        - eg. "often reported parameter values are tied to a model and its other parameters, so mixing values from different sources has its own caveats" (https://github.com/popsim-consortium/adding-species-manuscript/issues/7)
+The purpose of including a species into the stdpopsim catalogue
```
lauterbur: I think we want this part to be more general, not just referring to the stdpopsim catalog but making a good popgen simulation in general. Then we could use this paragraph as the introduction to section 6 (steps to add to the stdpopsim catalog), and edit the other stdpopsim references to reflect popgen simulations in general, and/or add references to section 6 below.

lauterbur: So instead we could start this section with something like "Purposes of generating population genetic simulations include testing and benchmarking inference methods, creating data for Approximate Bayesian Computation and for training machine learning methods, and <... should complement the introduction ...>. Inference and methods testing improve the more aspects of realism can be incorporated into the simulations, so the effects of linkage, recombination rate variation, demographic changes, and other real-world factors affecting patterns of genetic variation in real species can be modeled. So when choosing to generate realistic population genetic simulations for a species of interest, certain elements should be known. We do not attempt here to create tutorials for any specific population genetic simulation programs (such as ms, msprime, ..., or SLiM), but to describe the necessary elements for generating realistic simulations for a species of interest with any of these programs. Concretely, all species would ideally have:"

adding_species_manuscript.md
```diff
@@ -162,28 +164,113 @@ with stdpopsim.
...
+is one of the most important roles of stdpopsim.
+Ideally, each species will have at least one citeable **demographic model**
+(more than one because the appropriate demographic model to compare to a given dataset
+usually depends on the sampling location(s) of the data),
+although at minimum, stdpopsim requires a
+a plausible and citeable effective population size estimate,
+as a single-population simulation with a reasonable (effective) size
+should at least give comparable levels of overall genetic diversity.
+
+The goal of much population genetics work
```
lauterbur: "The above factors are sufficient for modeling neutral variation in populations, for example to infer past migration events or population size changes. However, the goal of much population genetics work..."

adding_species_manuscript.md
```diff
@@ -182,13 +185,36 @@ Users can also add other attributes of a species to the simulation that are not
...
+To consider this, it helps to consider the use cases for stdpopsim.
+If there is not a chromosome-level assembly,
+then we need to either just simulate many, small chunks of genome,
+or simulate an anonymous chromosome, not specific to the species.
+We could treat each of the many contigs of an assembly as a "chromosome" in stdpopsim.
+However,
+this is not right because it's missing the effects of linked selection (although it may be useful),
+but more crucially, it would be a significant amount of effort for stdpopsim developers
+that wouldn't lower barriers for downstream users much at all.
+With a great many contigs,
```
lauterbur: I think we could make this section more general by using this paragraph (without the reference to stdpopsim), and a little bit about what you'll be missing (linkage, so selection analyses won't be as good as they would otherwise be). Then the stuff above this about stdpopsim development etc. could go into section 5 (examples from hackathon/what we learned).

lauterbur:
Thanks, Peter! This is great. I've commented with some thoughts re: making it more general (following https://github.com/popsim-consortium/adding-species-manuscript/issues/1), and we can get into more specifics about stdpopsim in sections 5 and 6. What do you think?

adding_species_manuscript.md
petrelharp: Yes, good point. I've had a go at this, but more could be done.

PR #33, opened 2022-04-21
Latex mig

andrewkern:
moving the manuscript over to latex format here. this PR includes changes to the Makefile, the references, and adds a `.tex` file which should is _now_ the doc to add. I didn't take the liberty of deleting the now deprecated `.md` file, but we could do that as well if people like


adding_species_manuscript.tex
```diff
@@ -377,9 +387,9 @@ \subsubsection{Implementing a population genomic
 we hope that many more will be available in the near future. This is on
 the horizon\\
```
lauterbur: does this line break need to be here?

adding_species_manuscript.tex
```diff
@@ -0,0 +1,847 @@
...
+non-model species. Table 1 displays some considerations and solutions
+For example, if all the parameters above are available, except for a
+chromosome level assembly, then one might choose to include only the
+larger contigs.
+
+Table 1. Guide to missing parameters.
+
+% Please add the following required packages to your document preamble:
+% \usepackage{booktabs}
+\begin{table}[]
```
lauterbur, line 530: table needs linebreaks, but should these be added later when the content is finalized?

lauterbur (approved):

lauterbur:
Awesome, thanks for doing that! I'm happy with deleting the deprecated .md file.

adding_species_manuscript.tex
andrewkern: nope

adding_species_manuscript.tex
andrewkern: we can deal with these later when we have the content there

andrewkern:
okay @lauterbur I made these edits. I'm going to squash and merge my edits in and then folks can edit the tex file directly

Issue #31, opened 2022-04-20
Quantitative genetics/genomics simulations often require whole-chromosomes or whole-genomes


gregorgorjanc:
Quantitative genetics/genomics simulations often require whole-chromosomes or whole-genomes - can mention this in the section where we describe why would one want to simulate more than, say, a 1cM region.

Issue #30, opened 2022-04-20
Cite Earth Biogenome project in the conclusion 


gregorgorjanc:
Cite Earth Biogenome project in the conclusion - currently mentioned Darwin Tree of Life is just one of many such projects that all fall under the umbrella of the Earth Biogenome (as far as I know!)

PR #35, opened 2022-04-25
Addressing #30 and partially #31

gregorgorjanc:
This is a set of minor edits

### DIFF
--- a/adding_species_manuscript.tex
+++ b/adding_species_manuscript.tex
@@ -48,7 +48,7 @@
      }
 
 
-% Avoid pandoc bug when there are lists in the body. 
+% Avoid pandoc bug when there are lists in the body.
 \providecommand{\tightlist}{%
 \setlength{\itemsep}{0pt}\setlength{\parskip}{0pt}}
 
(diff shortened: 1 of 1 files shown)


adding_species_manuscript.tex
```diff
@@ -802,9 +803,9 @@ \section*{Conclusion}\label{conclusion}}
 easy inclusion of new species genomes, and community-maintained accuracy
 and correctness. We additionally briefly describe how to add species to
 the stdpopsim catalog, and how the quality control process for species
-inclusion works. Currently large-scale projects such as the Darwin Tree
-of Life (TODO: CITE) project are generating tens of thousands of genome
-assemblies-- each of these, with some prior knowledge of mutation and
+inclusion works. Currently large-scale efforts such as the Earth Biogenome
+and its affiliated project networks are generating tens of thousands of genome
```
petrelharp, line 807: I don't know the landscape here - so DTOL is an affiliate project of EBP? (should be "earth biogenome project", btw) Has anyone else actually produced lots of assemblies yet besides DTOL? If not it's worth singling out DTOL here (although mentioning EBP too?).

petrelharp:
looks good - minor comment/question?

adding_species_manuscript.tex
gregorgorjanc: True, DTOL is churning more out at the moment, but my understanding is that others are on that track. I am happy to go either way, but though that it would be good to point the general state of afairs.

lauterbur (approved):

PR #36, opened 2022-04-28
Phylofig

lauterbur:
added phylogenetic tree figure


lauterbur:
Looks like my commits got a little mixed up. The only thing to look at here is the phylogenetic tree figure, cover letter draft will come later.

PR #38, opened 2022-05-04
Adding authorship info

awohns:
I'm very happy to be a coauthor on this paper. I've added my name (I'm guessing you can't use my full name, so I put in my middle initial) and affiliation. I will also review the manuscript and QC a species.
Thanks very much all!


petrelharp:
Great! I've added this on top of #39 to make things simpler, so I'll close this one. And, I've put in your middle name - don't know of any reason we can't use it? - but let me know if you're rather have it just "W".

PR #40, opened 2022-05-07
Affiliation for @aabiddanda

aabiddanda:
Small PR, just shifted all other affiliations by 1 for inclusion. Should be tee'd up for squash and merge (two commits locally were due to fixing some conflicts). Addresses #32 partially.


petrelharp:
Awesome, thanks!

PR #45, opened 2022-06-06
Update adding_species_manuscript.tex for BosTau Ne

gregorgorjanc:
Updating description regarding the effective population size in the basic genome model and the demographic model.

@gtsambos and @petrelharp I would appreciate your revision of this change so that we are all on the same page.

Thanks!


petrelharp:
I like it - no complaints.

igronau:
Looks good to me. @lauterbur - can you merge?

lauterbur:
With pleasure.

gtsambos:
Sorry for the late reply all -- I had limited access to laptop internet while I was in CZ. This looks good to me, thanks for doing it @gregorgorjanc!

PR #48, opened 2022-07-13
Ilan's edits to description of cow model in the "Examples" section

igronau:
Also added a brief comment on the option of specifying a mutation rate in each demo model in the previous section


andrewkern (approved):

andrewkern:
okay @igronau I'm going to merge these and then do a pass

PR #57, opened 2022-08-08
add gene conversion paragraph

fbaumdicker:
I have added a short paragraph on the gene conversion feature.
Please consider this as a suggestion and feel free to improve this draft and change whatever seems necessary. :)


igronau (approved):
Looks good. This can be merged as is.

petrelharp:
I think the thing to do at this point is to manually extract this paragraph and insert it into the current doc - @igronau?

gphocs-dev:
I'll take care of that in my (final :-)) pass

petrelharp:
DId you get this in there, @igronau? If so, let's close this?

igronau:
I'm going to apply all of my edits once @lauterbur is done with hers. I'd like to keep this PR up to remind me to integrate this into my edits.

adding_species_manuscript.tex
```diff
@@ -212,7 +212,7 @@ \section*{Introduction}\label{introduction}}
 In general, the usefulness of population genomic simulations increases
 when they more realistically represent the species being simulated---that
 is, as they incorporate more relevant elements of the species' biology. Important
-elements include genomic features such as mutation and recombination
+elements include genomic features such as mutation, recombination, and gene conversion
```
igronau, line 217: I decided not to mention gene conversion in this context, because this sentence already contains many element.

adding_species_manuscript.tex
```diff
@@ -350,6 +350,8 @@ \section*{Introduction}\label{introduction}}
 reduce variance, generally, relative to independent simulations of equivalent genetic material.
 This has a particularly striking effect in long stretches of low recombination rates,
 as observed for instance on the long arm of human chromosome 22 \citep{Dawson2002}.
+In bacteria, a similar effect occurs due to genome-wide linkage that is broken only
+by gene conversion of short segments.
```
igronau: I added this in

adding_species_manuscript.tex
```diff
@@ -425,14 +427,30 @@ \section*{Introduction}\label{introduction}}
...
+which gives the mean of the geometric distributed length of the segment transferred.
+Similar to species-specific mutation and recombination rates,
+default values for the gene conversion rate and the gene conversion tract length can be included in the species catalog, 
+if they pass the quality control step.
+% FB: We might want to consider adding some more default gc parameters to the catalog. If we do so in time, we should mention that here.
+For example, the species-specific gene conversion parameters of \emph{Escherichia coli} have now been included in the catalog.
+However, the rate at which gene conversions occur in eukaryotes is often several times higher than 
+the single crossover recombination rate, sucht that simulating including gene conversion can take considerably longer.
+Therefore gene conversion is disabled by default in \stdpopsim,
+but can be easily included with the use\_species\_gene\_conversion flag using the parameters from the species catalog or custom parameters.
```
igronau, line 469: I added this text in, with a few edits. Please check in my next PR

references.bib
```diff
@@ -963,3 +963,17 @@ @article{Huber2018
...
+	year = {2017}
+}
```
igronau, line 979: added this ref

adding_species_manuscript.tex
```diff
@@ -653,7 +671,7 @@ \subsection*{Extracting model parameters from the literature}
 \subsection*{Filling out the missing pieces}
 
 For some species it may be difficult to obtain citeable values for the necessary model parameters, even when combining different sources. We provide several suggestions for dealing with this scenario (see Table \ref{tab:param-mod}).
-If the species of interest does not have citeable mutation or recombination rates, the average genome-wide rates published for a closely related species may be used. This may slightly skew the average genetic diversity and the average genetic linkage, but will still likely produce useful simulations. A similar approach can be applied to the DFE if simulating natural selection.
+If the species of interest does not have citeable mutation, recombination, or gene conversion rates, the average genome-wide rates published for a closely related species may be used. This may slightly skew the average genetic diversity and the average genetic linkage, but will still likely produce useful simulations. A similar approach can be applied to the DFE if simulating natural selection.
```
igronau, line 720: I'm considering sticking with the original wording because I'm not sure that we want to put gene conversion in the same basket as recombination and mutation rates here. Mutation and recombination rates are required for each model. We don't require parameters for gene conversion for all models. So, one might add a species model without gene conversion parameters. You can't do this for mutation and recombination rates.

adding_species_manuscript.tex
```diff
@@ -710,6 +728,11 @@ \subsection*{Filling out the missing pieces}
 			will affect the impact of selection, linkage, and linked selection
 			\\
 			\hline
+			Gene conversion parameters &
+			borrow from closest relative with a citeable rate &
+			will affect the impact of selection, linkage, and linked selection
+			\\
+			\hline
 			Demographic model &
```
igronau, line 783: I'll add this to the table, but will mention that this is relevant only if modeling gene conversion is central to the species evolution (such as in bacteria and archea).

igronau:
@fbaumdicker , I wrote some comments on this PR. Will apply these edits (modulo my comments above) in my next PR. Please let me know if you have any additional input.

adding_species_manuscript.tex
lauterbur: Would it be valid to use gene conversion rates from a closely related species? (Or at least as valid as for mutation and recombination rates.) If yes, I think it should be made clear here as well, but perhaps also state explicitly that this is only if a gene conversion rate is required to model the important diversity in your organism.

fbaumdicker: I think this would be a valid approach.

adding_species_manuscript.tex
fbaumdicker: We have meanwhile changed the way gene conversion is implemented in stdpopsim. See also https://github.com/popsim-consortium/stdpopsim/pull/1355
In particular, we now use a gene_conversion_fraction for eukaryotes and there is a bacterial_recombination_flag that uses the recombination_rate of the species as gene_conversion.
Thus I think we should update this here. In particular, this means that for bacteria a gene conversion rate has to be provided now (as recombination_rate) when adding a species to the catalog.

fbaumdicker: I would change the text myself, but can only do so starting next Thursday 09/22/22.

adding_species_manuscript.tex
igronau: I'll try to clarify this point here by adding a separate sentence about gene conversion here. I don't want to lump it together with mutation and recombination rates, which are basic essentials for every species.

igronau:
The text on gene conversion has considerably changed and expanded in the current version. @fbaumdicker - you should review this in the latest PR

adding_species_manuscript.tex
```diff
@@ -367,6 +367,9 @@ \section*{Introduction}\label{introduction}}
 generally reduce variance relative to independent simulations of equivalent genetic material.
 This has a particularly striking effect in long stretches of low recombination rates,
 as observed for instance on the long arm of human chromosome 22 \citep{Dawson2002}.
+In bacteria, a similar effect occurs due to genome-wide linkage that is broken only
+by gene conversion of short segments {[TODO: citation?]}.
+
```
fbaumdicker, line 372: This might be a suitable citation:
[10.1016/j.tim.2010.04.002](https://doi.org/10.1016%2Fj.tim.2010.04.002)

igronau: @fbaumdicker , I think I'm citing this reference in a similar context in the current version. BTW, all discussion about gene conversion has been significantly modified to better capture the new way that non-crossover recombination is modeled in stdpopsim. See the latest PR: https://github.com/popsim-consortium/adding-species-manuscript/pull/68. Let me know if you have any comments on current version.

fbaumdicker: Ah yes, thanks for the link to the newest PR. I have added one small comment to the latest PR. Besides that suggestion I think @petrelharp has already addressed all other points.

Issue #47, opened 2022-07-12
Specific writing and review tasks for paper


igronau:
I'm listing some action items with suggestions of possible authors who could address them. Feel free to pass on to a more suitable author, or take on a different task.

**Section: Standardized genome-wide simulations**
- [x] review the 1st paragraph in this section, which was rewritten to provide a fundamental overview of popgen simulations. (@petrelharp , @lauterbur )
- [ ] check that the main arguments about whole-chromosome simulations are clearly covered in 4th paragraph (@andrewkern , @petrelharp )
- [x] elaborate on example of Hsieh, et al. in the 5th paragraph (@RyanGutenkunst )

**Section: Updates to stdpopsim and the species catalog**
- [x] summarize key updates in msprime v1.0 in 1-2 sentences (@jeromekelleher , @FranzBaumdicker)
- [x] summarize model of gene conversion in stdpopsim in 5-7 sentences (@FranzBaumdicker)
- [x] add 5-7 sentences on uploading annotations, DFEs and mention selection manuscript (@petrelharp , @andrewkern )
- [x] fill in the three missing numbers in 2nd paragraph: (@lauterbur )
        * number of intro to stdpopsim workshops
        * number of hackathon participants
        * number of species added *after* the hackathon (and not immediately after)
- [ ] update fig 1 using updated catalog and mark in diff color species added during the hackathon (@lauterbur )

**Section: Guidelines for implementing a population genomic simulation**
- [ ] overall review of this section, which was quite extensively edited in this round (@lauterbur , @DanSchrider, @petrelharp ) 
- [ ] do we want to explicitly specify the max number of scaffolds we allow in species added to the catalog? This is mentioned twice in this section: beginning and end (@petrelharp , @jeromekelleher )
- [ ] does stdpopsim strictly require us to provide a generation time in every species? (@jeromekelleher , @petrelharp )
- [x] review paragraph on modeling selection in simulations (@andrewkern )
- [x] add citation(s) about usefulness of realistic simulations of selection (@andrewkern , @BenHaller)
- [ ] add citation to selection paper (preprint?)
- [x] add citations to studies that infer DFEs (try to cover a few species) (@KirkLohmueller )
- [x] write short paragraphs on liftovers (@andrewkern )
- [x] figure out an appropriate citation for estimating $\theta$ from $\pi$ (currently, we're citing Tajima `89) ( @andrewkern  )
[...]

PR #58, opened 2022-08-09
MEL edits

lauterbur:
Addressing tasks at https://github.com/popsim-consortium/adding-species-manuscript/issues/47 and general edits.


lauterbur:
I haven't yet merged @fbaumdicker 's gene conversion section, I can do these both together if we're okay with them. @gphocs-dev what do you think?

gphocs-dev (approved):
looks good and can be merged, together with Franz's edits. What about the catalog figure? does this need to be updated?

PR #63, opened 2022-08-26
fix previous bad commit

lauterbur:
figure pdf is back, R code merge detritus removed, cleaned up R code in general


figs/species_fig.R
```diff
@@ -225,14 +202,9 @@ tree_plot +
...
   scale_color_manual(name="count", na.value="white", values = c("grey","black")) +
-  # scale_x_discrete(position="top") +
   ggplot2::xlim(0, 8000) +
   ggplot2::ylim(-1.5,21.5)
 
 ggsave("./species_fig_temp6.png",height=9,width=9)
 
 # increase size of column labels, angle, remove "NA" in legend - do this outside R because ggtree doesn't want to
-=======
-dev.off()
```
petrelharp, line 236: This is necessary to close the device opened by the call to `pdf( )` way up there... but, I'm not sure how the figure's being made any more?

petrelharp:
Does running this script produce the pdf file? (`species_fig.pdf`, I think?)

petrelharp:
We should probably merge this in, @lauterbur?

lauterbur:
Edits incorporated in https://github.com/popsim-consortium/adding-species-manuscript/pull/66

lauterbur:
Ah, so since the packages I'm using to produce the base figure don't
*quite* do everything (adjusting the column label height without screwing
up the justification terribly, for example), I need to do some manual
manipulation after the figure is generated. I had left that as a png
(produced as "species_fig_temp6.png") and then converted to pdf in gimp
when I did the tweaks, but I can generate it as a pdf instead if that's
preferred.

PR #66, opened 2022-09-12
Mel lastpass

lauterbur:


lauterbur:
Okay, in theory this is my last pass, editing baton goes to @gphocs-dev !

I've merged the figure PR here (https://github.com/popsim-consortium/adding-species-manuscript/pull/63) and incorporated Jerome's comments (rather than merging his PR directly). I've left a few comments/questions in the text. If we want the simulation model figure suggested for the mosquito example I'll make it up quickly, just let me know.

Jerome made a good point that the _Anopheles_ example needs some work. I've done some editing, but I see that there was a lot of back and forth in the QC process (https://github.com/popsim-consortium/stdpopsim/pull/906) - I think it would be valuable to address some of this and some of the potential consequences (as the Bos taurus example does in its last paragraph), but that would require more substantial rewriting for that section. Thoughts?

I restructured the conclusion a bit to better fit the current structure of the paper and abstract/intro. Minor notes: American vs British spelling should be consistent one way or the other; "population genetic" vs "population genomic" should be consistent as appropriate.

PR #68, opened 2022-09-20
Ilans final edits

igronau:
I'm just pushing my latest edits now. Will follow up with a more detailed description of what I did and the final minor tasks left. Stay tuned...


igronau:
Summary of main changes: (highlighted in **bold are some action items**)
1. A few minor additions to abstract. **I'd like others to review the new abstract** to make sure it conveys the main points we wish to make in the paper
2. Intro: last paragraph was edited to provide a more comprehensive review of the paper, while still focusing on the "guidelines" section
3. The utility of stdpopsim section: this section was a  bit scattered, with rough transitions between paragraphs. I made some minor adjustments to improve the flow between the three paragraphs and focus on the main ideas.
4. Slight graphical modifications of fig 1 (thanks, @lauterbur ).
5. The paragraph on gene conversion in the "additions to stdpopsim" was completely rewritten to describe the new way we parameterize GC. **@petrelharp and Franz should review this to make sure I got everything correct.** There are also some potential updates to the EscCol model that may require some additional tweaks here. See my comments [to this PR](https://github.com/popsim-consortium/stdpopsim/pull/1355).
6. Added references for various issues having to do with mis-specification of the demographic model (in the "Basic setup" subsection). I tried, but couldn't find additional references for the main point that mis-specifying the demographic model may lead to unrealistic patterns of variation. We're currently using [Navascues and Emerson (2009)](https://pubmed.ncbi.nlm.nih.gov/19735451/), but @petrelharp rightfully suggested that there likely are other/better references. **Do any good references come to mind?**
7. In the "missing pieces" subsection, I edited table 1 to try to clarify the last column. Not sure how good it is. **This definitely deserves a good review**. Following @lauterbur's edits, we now don't mention any of this in the main text and it's only described in the table. I think it works nicely, as long as we get the table to be 100% clear. I also followed @jeromekelleher's suggestions and streamlined the 2nd and 3rd paragraphs in that subsection to drive the main message.
8. The AnoGam example. @lauterbur is creating a figure to add here based on what we have in the docs. I also made some edits to emphasize the lessons we can learn from this example. I recall people suggesting that there is more we can mention here about the QC, but I couldn't find anything worth mentioning from the QC issues. **If you remember anything specific, let me know (@andrewkern? @petrelharp?).**
9. I didn't yet make a thorough pass over the conclusions, but they look quite good in the current state, so I don't think I'll make any significant changes there.
10. Added a short acknowledgement statement to workshop participants. **Please review this.** We also need to make sure that all funding is mentioned in the last section.

petrelharp:
[adding_species_manuscript-diffmain.pdf](https://github.com/popsim-consortium/adding-species-manuscript/files/9608931/adding_species_manuscript-diffmain.pdf)
Here's a latexdiff of the changes. For the record, to do this I did:
```
git checkout main
git pull origin
git fetch origin pull/68/head:ilan_pr
git checkout ilan_pr
latexdiff-git -r main adding_species_manuscript.tex
pdflatex adding_species_manuscript.tex
bibtex adding_species_manuscript.aux
pdflatex adding_species_manuscript.tex
pdflatex adding_species_manuscript.tex
```

adding_species_manuscript.tex
```diff
@@ -453,34 +421,65 @@ \subsection*{Additions to \stdpopsim}
...
-which is essential for organisms such as \textit{E.\ coli}
-that lack crossing over.
-\colorbox{yellow}{[ TODO: summarize model of GC (see PR\#57) ]}
+Expanding the species catalog required adding several functionalities to the simulation framework.
+We thus upgraded the neutral simulation engine, \texttt{msprime}, from version 0.7.4 to version 1.0. \citep{Baumdicker2022}.
+This upgrade provides a number of benefits such as a discrete site model of mutation,
+so that simulated data will now %, as in real data,
+have a small proportion of sites with multiple mutations and possibly more than two alleles.
+Another key feature enabled by this upgrade was recombination by gene conversion,
+which is essential for modeling genomes of bacteria and archea.
```
petrelharp: My understanding is that what bacteria do acts like gene conversion (and so is implemented using the "gene conversion" widgets of the simulators) but is called something different (e.g. "horizontal transfer"), so this ought to be readjusted a bit?

adding_species_manuscript.tex
```diff
@@ -453,34 +421,65 @@ \subsection*{Additions to \stdpopsim}
...
+Gene conversion affects shorter segments than crossover recombination and creates distinct patterns of genetic diversity along the genome~\citep{Korunes2017}.
+In bacteria and archaea, homologous recombination occurs primarily through gene conversion rather than single crossover recombinations.
+As a result, such species cannot be realistically simulated with a recombination model that considers only crossovers.
+Gene conversion in such species is implemented in \stdpopsim by setting the
+\texttt{bacterial\_recombination} flag to \texttt{True} (as a feature of the genome)
+and the  \texttt{gene\_conversion\_length} parameter to the average gene conversion tract length.
```
petrelharp: I don't think we need to put in the actual API options?

petrelharp: Hm, also - the gene conversion option works in SLiM now also (where the tract length is not geometric but negative binomial), but this paragraph is headed by 'msprime'.

adding_species_manuscript.tex
```diff
@@ -453,34 +421,65 @@ \subsection*{Additions to \stdpopsim}
...
+This will result in all recombinations being simulated via gene conversion
+(no crossover recombinations), where the tract length is sampled from a geometric
+distribution whose mean is the specified length.
+For example, the model for \textit{Escherichia coli} has been updated in the \stdpopsim catalog to have a gene conversion rate of $8.9\times 10^{-11}$ with an average tract length of 345 bases \citep{Wielgoss2011,Didelot2012}.
```
petrelharp: actually it has a *recombination rate* of this, but the "bacterial recombination" flag is on, which means that "recombination" is interpreted as gene conversion

adding_species_manuscript.tex
```diff
@@ -453,34 +421,65 @@ \subsection*{Additions to \stdpopsim}
...
 % Elise: Not sure the blow fits in the GC paragraph?
-In addition, the update to msprime version 1.0~\citep{Baumdicker2022}
-provides a number of additional benefits such as a discrete site model of mutation,
-so that simulated data will now, as in real data,
-have a small proportion of sites with multiple mutations and more than two alleles.
-
-Moreover, we have extended \stdpopsim so that
-genome annotations can be associated with a genome assembly.
-These can be used to simulate selection at a subset of sites (e.g., the annotated coding regions)
+% Ilan: the idea is that the model of GC is enabled due to the upgrade to msprime 1.0. I made sure to mention this above. This is still a bit tacked-on here, but still probably the best place to mention this.
```
petrelharp: I think we should mention here that GC is *not* on by default because it makes things significantly slower/bigger,.

adding_species_manuscript.tex
petrelharp: so - I think GC should be a new paragraph

adding_species_manuscript.tex
```diff
@@ -525,53 +523,59 @@ \subsection*{Basic setup for chromosome-level simulations}
...
   At minimum, an average genome-wide recombination rate needs to be specified, which is typically available for well assembled genomes.
-  %As with the average mutation rate, if an estimate is not available for your species of interest, you may use an estimate obtained for another, closely related, species.
+  %ILAN: added this part about GC
+  Recall that for bacteria and archea, which primarily experience recombination by gene conversion,
+  the recombination rate corresponds to the rate of gene conversion,
+  and the average tract length should also be specified
+  (see details in previous section).
+  If one wishes to model gene conversion together with crossover recombination,
+  then they should specify the fraction of recombinations done by gene conversion
+  as well as the average tract length (per chromosome).
```
petrelharp, line 506: People can still add GC rates themselves to contigs for which we haven't specified GC rates, but this is no longer part of the standard API - perhaps this should say instead just something saying that GC is available but optional.

adding_species_manuscript.tex
```diff
@@ -634,114 +641,112 @@ \subsection*{Extracting parameters from the literature}
...
+The table also mentions the main discrepancies between the simulated data and real genomic data,
+which can be caused by mis-specification of each parameter.
+
+%ILAN: shortened the paragraph by removing a lot of the low-level considerations,
+%  following a suggestion by Jerome
+Several researchers who participated in the ``Growing the Zoo'' hackathon wished to add species
+whose genome assemblies are composed of many relatively small contigs,
+unanchored to chromosome-level scaffolds.
+Although we wish to keep \stdpopsim as inclusive as possible,
+we made a conscience decision to only add species with chromosome-level assemblies.
```
petrelharp: It's not so clear what 'inclusive' means here (like, we also aim to be an inclusive community of people working together); how about
```suggestion
Although we had not previously put restrictions on which species might be added,
we decided that we would only add species with chromosome-level assemblies.
```

adding_species_manuscript.tex
```diff
@@ -634,114 +641,112 @@ \subsection*{Extracting parameters from the literature}
...
+downstream users of these models.
+
+ 
+That being said, simulation is still possible and potentially useful for species with partial genome builds.
 One way to deal with this situation is to include only the longer contigs or scaffolds,
 treating them as separate chromosomes in the simulation.
 Some of these contigs will map to the same chromosome, 
 so simulating them separately will not capture the genetic linkage between them.
 However, this provides a reasonable approximation for many purposes, at least for genomic regions far from the contig edges.
 Short contigs can either be omitted from simulation, or lumped together into one (or several) longer pseudo-chromosome(s).
```
petrelharp: I think the deleted stuff in the rest of this paragraph is useful, and it's part of the point of this paper to explain the underlying tricky issues.

petrelharp: The point about "false precision" is that if someone thinks "oh, I'm precisely simulating segment 18374499 to 4885722 of chromsome 18" then they're missing the point that well yes the features on that segment might be 'precise', but the linked effects of other bits of chromosome 18 are missing.

adding_species_manuscript.tex
```diff
@@ -634,114 +641,112 @@ \subsection*{Extracting parameters from the literature}
...
-			will affect the impact of selection, linkage, and linked selection
-			\\
-			\hline
-			Demographic model &
-			at least Ne is required and is estimable from mutation rate and genetic data     &
-			the demographic history (e.g. bottlenecks, expansions, and population splits and migration) affects patterns of variation substantially {[}CITE{]}, a constant Ne is not ideal \\
-			\hline
-	\end{tabular}
-\end{table}
+% Ilan: shortened this last part and removed mention of the option to create anonymous chromosomes. IMO, for many purposes, it's probably better to keep the sites independent rather than create aritificial and erroneous linkage between them.
```
petrelharp: I don't agree. By having a realistic amount of linkage you, for instance, would end up with a realistic amount of background selection acting across the genome (and so a simulation of unlinked loci would end up with too much genetic diversity). The precise landscape of background selection intensity would not be right, but the general statistical properties of it should be right-ish, to the extent that a random stitching together of contigs looks like a real chromosome.

petrelharp:
Generally, looks great! I've made comments in a few places. Thanks a bunch!

adding_species_manuscript.tex
igronau: I'm editing this now. My plan is to start by discussing "non-crossover recombination" without leaning on any specific terminology (such as GC or HGT). I'm adding citation to several studies on recombination in bacteria and archea

adding_species_manuscript.tex
igronau: I'll remove the mention of specific flags and parameter names, but will probably still say that there is a flag that tells the engine to use a model of non-crossover recombination. I'm also differentiating this discussion from the upgrade of msprime to v1.0 (but keeping it in the same paragraph ) and explicitly mentioning SLiM to clarify that the same model can be simulated with both engines

adding_species_manuscript.tex
igronau: This is what I meant, but I'll use the correct terminoligy

adding_species_manuscript.tex
igronau: I didn't get your point here. I want to stick to the framework as we describe it in the gene conversion paragraph. One may model GC by specifying for each chrom a % of recs done by GC + an average tract length. Do we want to say anything more here?

adding_species_manuscript.tex
igronau: accepted your suggestion and will include in the next PR

igronau:
I pushed a version which addresses @petrelharp's comments regarding the text on gene conversion. I now call it non-crossover recombination throughout the manuscript, unless we explicitly refer to gene conversion (in eukaryotes). I'll continue with other comments and edits later.

adding_species_manuscript.tex
```diff
@@ -155,28 +155,33 @@ \section*{Abstract}
...
 to confidently answer a given question, or what information is required
 to produce simulations of that desired realism.
-\Stdpopsim is a community-developed tool that seeks to lower this barrier
+\Stdpopsim is a community-developed framework that seeks to lower this barrier
 by making it easy to simulate complex population genetic models using
 up-to-date information.
-The initial version of the \stdpopsim catalog contained information for 
-6 species, most of which are well-characterized model organisms.
-Here, we report on community-driven efforts to expand the catalog 
+The initial version of \stdpopsim, the species catalog contained information for
```
jeromekelleher: Missing an "In" at the start of this sentence? The current version makes it sound like stdpopsim is just the catalog.

adding_species_manuscript.tex
```diff
@@ -317,82 +320,63 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
-This considerably increases the reliability of the resulting simulations in any downstream analysis.
-
-
-% \paragraph*{\textbf{Species-specific genomic architecture is challenging to model}}
-The goal of complete chromosome simulation is important for a number of reasons.
-The organization of genes on chromosomes is a key feature of a species' genome,
-and one that has largely been ignored in population genomic simulation (see \cite{schrider2020background} for a notable exception).
+Importantly, this increases the reliability of the resulting simulations in any downstream analysis.
+
+Another central goal of \stdpopsim is to promote whole-genome simulations,
```
jeromekelleher: s/central/important ?

jeromekelleher (commented):
LGTM, a few minor suggestions above

adding_species_manuscript.tex
igronau: Yes, that's a typo. I'll also try to reword that part of the abstract to better convey that the catalog is only one aspect of stdpopsim

adding_species_manuscript.tex
igronau: I think i wanted to avoid important because it was used in the previous sentence. Will think of a way to bypass this.

lauterbur: To be honest, I'm not a fan of "importantly" as a modifier in general, but that feels like nitpicking. ;-) We could just remove it, ie. "This increases the reliability..." and use "important" in place of "central" in the next sentence.

jeromekelleher: I like Elise's idea

igronau: This paragraph has two other 'important's in it. I'll use my thesaurus to replace one of them ;-)

adding_species_manuscript.tex
igronau: Done

adding_species_manuscript.tex
igronau: @petrelharp I'm still not quite sure what the point of 'false precision' is. I'm going to try to provide more insight here by presenting the pros and cons of lumping more clearly:

> When lumping short contigs together, one should weigh the benefits of fitting the amount of data of real genomes and capturing the linkage structure within each contig, against the adverse effects of inducing false or inaccurate correlation between variants from different contigs.

(in bold the modified part of the sentence)

adding_species_manuscript.tex
igronau: Okay. I'm making some more edits to this paragraph and will try to make this point in the end. My idea is to now describe cases with partial genome builds like this:
1) Treat long contigs as separate chromosomes [ main caveat is that you're missing out on linkage between contigs that map to same chrom ]
2) Lump together shorter contigs or omit them completely [ benefits and caveats to lumping ]
3) If the species 'genome build' only consists of short contigs, you may simulate artificial anonymous chroms. This (again) models linkage within each contigs and (approximately) captures some basic features such as BGS
4) Finally, you have the option of simulating independent sites, that you can do even without any genome build. however, then you miss out on the cool benefits of whole-chrom simulations, which is what we're trying to promote here

igronau:
Pushed some more edits incorporating comments from @petrelharp , @jeromekelleher , @lauterbur.
LatexDiff attached: [adding_species_manuscript-diff.pdf](https://github.com/popsim-consortium/adding-species-manuscript/files/9622765/adding_species_manuscript-diff.pdf)

adding_species_manuscript.tex
petrelharp: from slack: maybe we need an example or two in this section; like "if you want to fit the distribution of LD values then it's better to simulate a whole anonymous genome; while if you want to compare the LD values for a specific piece of genome to simulation then it might be better to simulate that specific contig, either unlinked or linked to other random contigs (and take several draws of which random contigs it's linked to)"

adding_species_manuscript.tex
petrelharp: okay, here's the point: suppose that someone wants to simulate a genome that is just in a lot of contigs; and they put together the code to run neutral simulations of lots of unlinked chunks of exactly the same size as their contigs, so that each bit of real data is carefully matched to a simulated chunk. Although this seems very precise and accurate somehow, there is (almost) no point at all to doing this careful matching up. The advantages to realistic matching up of chromosomes are, I think mostly: (1) realistic recombination maps and (2) realistic varying strengths of selection (with exon density); but you get neither advantage from lots of anonymous unlinked contigs.

This is not a hypothetical situation, and is tricky and something we spent a good deal of time talking about, which is why I'm trying to explain it.

igronau: I'm starting to get the point (finally). So, I plan to add a thoughtful discussion about the benefits of lumping contigs into pseudo chromosomes (mentioning the obvious caveats), using the arguments you make in your comments (here and below). You've now got me on board on the position that the benefits outweigh the negative effects :-). If I'm not mistaken, the same points apply here (when we discuss a build that has long and short contigs) and in the case we mention later (with only short contigs). So, I think it makes sense to address this once and not as two separate issues. I will also remove the mention of the option to omit the short contigs. That's always an option I guess, and it distracts from the main point here. Does this plan make sense, @petrelharp ?

igronau: BTW, we might not want to make strict distinction between short and long contigs here. Clearly, lumping makes more sense for shorter contigs, but it might make more sense to blur the line here a bit.

igronau:
This latest commit contains some additional edits to the paragraph on partial genome builds in the "missing pieces" section, following discussion with @petrelharp. Let me know if you have any more comments or suggestions regarding this delicate issue

adding_species_manuscript.tex
fbaumdicker: I think adding a subsection for gc in eukaryotes might be a good way to disentangle it here.

E.g. starting from Line 553:

\textbf{gene conversion (optional)}
If one wishes to model gene conversion in eukaryotes either together with crossover recombination or as a stand-alone process,
  then they should specify the fraction of recombinations done by gene conversion
  as well as the average tract length (per chromosome).

adding_species_manuscript.tex
```diff
@@ -155,42 +155,43 @@ \section*{Abstract}
...
-simulation model, including the required input data, common pitfalls and major considerations,
-and describe how new species models can be integrated into \stdpopsim.
-We also introduce several major advances to the realism of \stdpopsim's simulation ability,
-including gene conversion and provision of species-specific genomic annotations.
-Together, these advances to \stdpopsim will strengthen efforts to use and develop
-simulation-based population genomic inference methods, with particular advances
-for non-model organisms, making them available, transparent, and accessible to everyone.
+The initial version of \stdpopsim focused on establishing this framework using 
+six well-characterized model species.
+Here, we report on updates made in the new release of \stdpopsim (version 0.2),
```
lauterbur: call them "major updates"?

adding_species_manuscript.tex
```diff
@@ -155,42 +155,43 @@ \section*{Abstract}
...
+which include substantial additions to simulation capabilities and a significant expansion of the species catalog.
```
lauterbur: I suggest reordering, so "adding species" comes first, and more directly link why the simulation additions are important for adding species: eg. "which includes a significant expansion of the species catalog and substantial additions to simulation capabilities that make it possible to simulate a broader range of species."

adding_species_manuscript.tex
```diff
@@ -155,42 +155,43 @@ \section*{Abstract}
...
+%Ilan: moved this up here, to leave the end of the abstract to the bigger ideas
+Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
+Expansion of the species catalog involved community-driven efforts to increase coverage more broadly across the tree of life
+to 21 species, with 25 demographic models and 37 genetic maps.
+Our experience throughout this process was that people are indeed keen to put in the time and effort to include their study species,
```
lauterbur: rm "indeed"

adding_species_manuscript.tex
```diff
@@ -155,42 +155,43 @@ \section*{Abstract}
...
-simulation-based population genomic inference methods, with particular advances
-for non-model organisms, making them available, transparent, and accessible to everyone.
+The initial version of \stdpopsim focused on establishing this framework using 
+six well-characterized model species.
+Here, we report on updates made in the new release of \stdpopsim (version 0.2),
+which include substantial additions to simulation capabilities and a significant expansion of the species catalog.
+%Ilan: moved this up here, to leave the end of the abstract to the bigger ideas
+Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
+Expansion of the species catalog involved community-driven efforts to increase coverage more broadly across the tree of life
+to 21 species, with 25 demographic models and 37 genetic maps.
```
lauterbur: Do we need the specific numbers here? We have the details later, and I think the numbers throw off the flow here. If we want some idea of magnitude, perhaps "Through community-driven efforts, we expanded the number of species in the catalog by three-fold, which broadened species coverage across the tree of life." (Could add "..., and increased the representation of species with specific demographic models and genetic maps." but that's getting back toward throwing off the flow.)

adding_species_manuscript.tex
```diff
@@ -155,42 +155,43 @@ \section*{Abstract}
...
+Our experience throughout this process was that people are indeed keen to put in the time and effort to include their study species,
+but that simple, clear guidance is vital.
+Our intention with this paper is in part to provide another learning
```
lauterbur: For some reason I think "in part" doesn't work right here, perhaps "Consequently, our intention with this paper is also to provide another learning modality to meet that need, ..."?

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
 The initial release of \stdpopsim was
 restricted to only six well-characterized model species, such as
 \emph{Drosophila melanogaster} and \emph{Homo sapiens},
-but feedback from \stdpopsim workshops identified a widespread desire
+but feedback we received from the community identified a widespread desire
```
lauterbur, line 249: "feedback we received from the community identified" -> "community feedback displayed"

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
 to simulate a wider range of non-model species,
 and ideally to incorporate these into the \stdpopsim catalog for future use.
-That feedback, and subsequent efforts to expand the catalog, 
+This feedback, and subsequent efforts to expand the catalog, 
 also uncovered the need for a better understanding of when it is practical to create a realistic
```
lauterbur: I'm tempted to suggest "vital need" since this was apparently a Big Deal that we didn't originally realize was a Big Deal, but I can hear my high school English teacher tutting and citing Hemingway at me.

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+%the broader population genetics community.
+
+This paper reports on the updates made in the current release of \stdpopsim  (version 0.2),
+and is also intended as a resource for any researcher
+who wishes to develop whole-genome simulations for their own species of interest.
+%and maybe add them to the \stdpopsim catalog.
+%Ilan: I added a more comprehensive walkthrough the different sections
+We start by describing the main idea behind the standardized simulation framework
+of \stdpopsim,
+and then outline the main updates made to the \stdpopsim catalog and simulation framework
```
lauterbur, line 261: add "that have made the broad expansion of the species catalog possible" or similar? since it seems mostly agreed that the new things we discuss are those relevant to the catalog.

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+in the past two years.
+We then devote a major section of the paper to provide guidelines for
```
lauterbur: rm "devote a major section of the paper to "

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+generating population genomic simulations, either for the purpose of using them in one specific study,
```
lauterbur, line 264: "generating population genomic simulations of a specific species, either for the purpose of using them in one or more specific studies,"

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+or with the intent of adding these simulations to \stdpopsim.
```
lauterbur: "or with the intent of making this species model available for future work by adding it to \stdpopsim."

If you don't want to go with that, I'd sub "these simulations" with "this species model" since it's not the simulations themselves that are added

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+Among other things, we discuss when a whole-genome simulation is more useful than
```
lauterbur: "things" -> "considerations" ?

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
-using any simulator. We pay particular attention to the ways
-in which \stdpopsim eases this burden, and describe how new users might
-add their own species information to \stdpopsim.
-The latter is discussed in the \nameref{sec:examples} section, where we lay out in
-detail the simple process of incorporating the information discussed
-in the \nameref{sec:sim-guidelines} section into \stdpopsim.
+We specify the required input data,
+% (genome assembly, mutation and recombination rates, and demographic model),
+mention common pitfalls in choosing appropriate parameters,
+and suggested courses of action for species that are missing estimates of some necessary inputs.
```
lauterbur: "suggested" -> "suggest"

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+We conclude with examples from a couple of species recently added to \stdpopsim,
```
lauterbur: "a couple of" -> "two"

adding_species_manuscript.tex
```diff
@@ -248,38 +249,40 @@ \section*{Introduction}
...
+which demonstrate some of the main considerations involved in the process of designing realistic whole-genome simulations.
+While the guidelines provided in this paper are intended for any researcher interested in implementing a population genomic simulation using any software,
+we do highlight the ways in which the framework set up by \stdpopsim eases the burden involved in this process.
```
lauterbur: "we do highlight" -> "we highlight"

lauterbur: "the framework set up by \stdpopsim" -> "the \stdpopsim framework"

lauterbur: "... eases the burden involved in this process and facilitates reproducible research."

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
-% Then, another contributor independently creates a simulation model
-% based on the same literature sources and the documentation provided by the initial contributor.
-% The two separate models are then compared to each other by automated scripts.
-% Any discrepancies are resolved by the two contributors,
-Any discrepancies are resolved between the contributor and reviewer,
-if necessary with input of additional members of the community.
+and to make it more robust and more reproducible.
+Contributors collect parameter values for their species of interest from the literature,
+and then specify these parameters in a template file for the new model.
+%Ilan: I added some of the details commented out by Elise but streamlined it in one sentence
```
lauterbur: LGTM

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
+This model then goes through a vital peer-review process,
+which involves recreating the model based on the provided documentation,
```
lauterbur: "which involves another researcher independently recreating..."

I think it should be made clear that it's not just making sure the code runs, but actually another human going through the same thought process.

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
 effect on patterns of genome variation in myriad species
 \citep[e.g.,][]{McVicker2009,Charlesworth2012}. 
 In addition, the lengths of chromosome-scale shared haplotypes within and
-between populations provides valuable information.
-Methods that use such information, such as MSMC \citep{Schiffels2020},
-or IBDNe \citep{browning2015accurate} %, or the method of \citet{ringbauer2018estimating},
-perform best on long genomic segments, with realistic recombination rates.
+between populations provides valuable information on their demographic history.
+Demography inference methods that use such information,
+such as MSMC \citep{Schiffels2020}, or IBDNe \citep{browning2015accurate},
```
lauterbur: "or IBDNe" -> "and IBDNe"

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
+%, or the method of \citet{ringbauer2018estimating},
+perform best on long genomic segments with realistic recombination rates.
 Chromosome-scale simulations are clearly required to test (or, train) such methods,
```
lauterbur: "(or, train)" -> "(or train)"

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
 or to conduct power analyses for design of empirical studies that use them.
```
lauterbur: "for design of" -> "to design"

adding_species_manuscript.tex
```diff
@@ -317,133 +320,97 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
 \emph{Homo sapiens}, \emph{Pongo abelii}, \emph{Canis familiaris}, \emph{Drosophila melanogaster},
 \emph{Arabidopsis thaliana}, and \emph{Escherichia coli} (Figure \ref{fig:tree}).
 One way the catalog has expanded is through introduction of additional demographic models
 for \emph{Homo sapiens}, \emph{Pongo abelii}, \emph{Drosophila melanogaster},
 and \emph{Arabidopsis thaliana}, enabling a wider variety of simulations for these
-mostly model species.
-
-However,
-these species represent a small slice of the tree of life.
+thoroughly studied species.
```
lauterbur: "thoroughly" -> "well"

adding_species_manuscript.tex
```diff
@@ -453,34 +420,65 @@ \subsection*{Additions to \stdpopsim}
...
+		including the six species we published in the original release \citep[in blue]{Adrion2020}, and 15 species that have since been added (in orange).
+		Solid circles indicate species that have one (light grey) or more
+		(dark grey) demographic models and genetic maps.
+		Branch length were derived from the divergence times provided by TimeTree5 \citep{Kumar2022}.
+	\label{fig:tree} }
+\end{figure}
+
+Expanding the species catalog required adding several functionalities to the simulation framework of \stdpopsim.
+Some features were added by upgrading the neutral simulation engine, \texttt{msprime}, from version 0.7.4 to version 1.0. \citep{Baumdicker2022}.
+Among other things, this upgrade includes a discrete site model of mutation,
```
lauterbur: "things" -> "benefits" or "features"

adding_species_manuscript.tex
```diff
@@ -453,34 +420,65 @@ \subsection*{Additions to \stdpopsim}
...
+which enables simulating sites with multiple mutations and possibly more than two alleles.
+Another key functionality added to the simulation framework of \stdpopsim was modeling of non-crossover recombination.
```
lauterbur: "the simulation framework of \stdpopsim" -> "\stdpopsim's simulation framework"

lauterbur: "modeling of non-crossover recombination" -> "modeling non-crossover recombination" or "the ability to model non-crossover recombination"

adding_species_manuscript.tex
```diff
@@ -489,24 +487,23 @@ \section*{Guidelines for implementing a population genomic simulation}
...
 \subsection*{Basic setup for chromosome-level simulations}
 
-
-To run a simulation requires a description of the organism's demography and mechanisms of genetic inheritance.
-Although in practice many parameters describing these processes might be only roughly guessed at,
-simulation software requires unforgivingly precise values.
-We start by describing how and where to find appropriate values, and some possible
-alternatives when values for the ideal parameters are not known.
+Implementing a realistic population genomic simulation for a species of interest
+requires a fairly detailed description of the organism's demography and mechanisms of genetic inheritance.
```
lauterbur: rm "fairly"

adding_species_manuscript.tex
```diff
@@ -525,53 +522,59 @@ \subsection*{Basic setup for chromosome-level simulations}
   \citep{Chakraborty2016} are likely to boost these initiatives. 
   When expanding the \stdpopsim catalog, we decided to focus on species with near-complete 
   chromosome-level genome assemblies (i.e., close to one contig per chromosome).
-  % \colorbox{yellow}{[TODO: do we want to specify a specific number here? (Jerome/Peter?)]} 
-  % MEL: I think this depends on if we're focusing on usefulness or storage as the constraint. If utility, a specific number wrt to "completeness" should be proportionate to the number of chromosomes. If storage, then it could be a straight maximum, but could result in situations with a small but poorly assembled genome being allowed (seems much more unlikely though). 
-  % PLR: I think the restriction is to be roughly the same as the number of chromosomes or at least chromosome arms; so how's this?
   This restriction was set mainly because species with less complete genome builds 
   typically do not have good estimates of recombination rate or genetic maps,
```
lauterbur: "or" -> "nor"

adding_species_manuscript.tex
```diff
@@ -525,53 +522,59 @@ \subsection*{Basic setup for chromosome-level simulations}
...
   \textbf{Recombination rates} (per generation per bp).
   Ideally, a population genomic simulation should make use of a chromosome-level \textbf{recombination map}, 
-  since the recombination rate is known to vary widely across chromosomes {Nachman2002},
+  since the recombination rate is known to vary widely across chromosomes \citep{Nachman2002},
   and this can strongly affect the patterns of linkage disequilibrium and shared haplotype lengths.
   When this information is not available, we suggest specifying an average recombination rate for each chromosome.
   At minimum, an average genome-wide recombination rate needs to be specified, which is typically available for well assembled genomes.
-  %As with the average mutation rate, if an estimate is not available for your species of interest, you may use an estimate obtained for another, closely related, species.
+  %ILAN: added this part about non-CO rec in bacteria
+  Recall that for bacteria and archea, which primarily experience non-crossover recombination,
```
lauterbur: rm "Recall that"

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+That being said, simulation is still potentially useful for species with partial genome builds.
+Longer contigs or scaffolds in these builds can be simulated separately and independently.
+%
+This approach allows us to model genetic linkage within each contig,
+but linkage between different contigs that map to the same chromosome will not be captured by the simulation.
+This provides a reasonable approximation for many purposes, at least for genomic regions far from the contig edges.
+%
+For shorter contigs, separate independent simulations will not be able to capture patterns of long-range linkage in a reasonably realistic way.
+%
+Thus, a viable option for shorter contigs is to lump them together into longer pseudo-chromosomes, trying to mimic the species' expected chromosome lengths.
```
lauterbur: "potentially viable"? since it depends on the question

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+%
+Despite their somewhat artificial construction,
+these pseudo-chromosomes have the important benefit of
+capturing patterns of linkage similar to those observed in real genomic chromosomes.
+%Ilan: tried to give a few concrete examples here
+If, for example, the main purpose of the simulation is to examine the distribution of lengths of shared haplotypes between individuals,
```
lauterbur, line 682: is "between individuals" redundant here?

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+or study patterns of background selection,
+then it makes sense to simulate such pseudo chromosomes.
+Note, however, that specific genetic correlations between different contigs lumped together will potentially be highly inaccurate.
```
lauterbur: "will potentially be" -> "can be"

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+So, if the main purpose of the simulation is to examine local patterns of genetic variation in certain loci of interest, then it might make more sense to simulate the relevant contigs separately (even if they are short), or to randomly sample several mappings of contigs to pseudo-chromosomes.
```
lauterbur: rm "So"
rm "certain"

lauterbur: "might make more sense" -> "would be more appropriate"

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+%
+Ultimately, the recommended mode of simulation for species with partial genome builds depends on the intended use of the simulated genomes.
```
lauterbur: I think this should be the last sentence of the paragraph/section since it's the real take-home message

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
+%Ilan: I moved this up here because this was the most straightforward step
+% I saw in the qc pr that there was some discussion about recombination rates
+% for the x chrom. Anything worth reporting? Maybe the pending estimates from 
+% the paper from Omar's group?
+Estimates of average \textbf{recombination rates} for each of the chromosomes (excluding the mitochondrial genome)
+were taken from a recombination map inferred by \citet{Pombi2006} which itself included information from
+\citet{zheng1996integrated} (Figure \ref{fig:anogam}A).
+As direct estimates of \textbf{mutation rate} (e.g., via mutation accumulation) do not currently exist for \emph{Anopheles gambiae},
 we used the genome-wide average mutation rate of $\mu=3.5 \times 10^{-9}$ mutations per generation per site,
 estimated for \textit{D.~melanogaster} by \cite{Keightley2009}
```
lauterbur: "estimated for its relative D. melanogaster" to reinforce the point that they didn't just choose any old estimate

adding_species_manuscript.tex
```diff
@@ -634,175 +640,205 @@ \subsection*{Extracting parameters from the literature}
...
 on the AgamP4 \textbf{genome assembly} \citep{Sharakhova2007}, which 
 was downloaded from Ensembl \citep{ensembl2021} via \stdpopsim's
 utilities that interact with Ensembl. These utilities
 make it easy to accurately retrieve basic genome information and construct the appropriate Python data structures.
 
-As direct estimates of \textbf{mutation rate} (e.g., via mutation accumulation) do not exist for \emph{Anopheles gambiae},
+%Ilan: I moved this up here because this was the most straightforward step
+% I saw in the qc pr that there was some discussion about recombination rates
+% for the x chrom. Anything worth reporting? Maybe the pending estimates from 
+% the paper from Omar's group?
```
lauterbur, line 728: I think this (recomb. rate on X) is what I was thinking about too. (For instance the resolution about using a recombination calculated at the middle of X vs calculated using 90% of X: https://github.com/popsim-consortium/stdpopsim/pull/906#issuecomment-1136160294)

lauterbur: Which leads me to a further thought - in the paper these are given as cM/Mb, should we point out that there can be necessary conversion calculations when papers use different units? (Because this is a place things could go wrong.)

adding_species_manuscript.tex
```diff
@@ -852,159 +890,86 @@ \subsection*{\texorpdfstring{\emph{Bos taurus} (cattle)}{Bos taurus (cattle)}}
...
+The issue of fitting the rates used in simulation with those assumed during inference was discussed during the independent review of this demographic model, and it raised an important question about recombination rates. Since \cite{MacLeod2013} use runs of homozygosity to infer the demographic model, their results depends on the assumed recombination rate. The recombination rate assumed in inference (\(r=10^{-8}\)) is \(8\%\) higher than the one used in the species model (\(r=9.26\times 10^{-9}\)). In its current version, \stdpopsim does not allow specification of a separate recombination rate for each demographic model, so we had no simple way to adjust for this. Future versions of \stdpopsim will enable such flexibility. Thus, we note that genomes simulated under the demographic model as currently implemented in \stdpopsim 
+might have slightly higher linkage disequilibrium than observed in real cattle genomes.
 However, we anticipate that this would affect patterns less
 than selection due to domestication and selective breeding,
-which are not modeled here.
+which are as of yet not modeled at all in the simulations of \stdpopsim.
```
lauterbur: "in the simulations of \stdpopsim" -> "in \stdpopsim simulations"

adding_species_manuscript.tex
```diff
@@ -852,159 +890,86 @@ \subsection*{\texorpdfstring{\emph{Bos taurus} (cattle)}{Bos taurus (cattle)}}
...
 
-We also present basic considerations for implementing
-population genomic simulations, agnostic to simulation software, based on
-insights from the community-driven process of expanding the \stdpopsim catalog. 
-We describe the steps of determining if a species-specific population
-genomic simulation is appropriate for the species and question, what
-data is necessary and why, special considerations for finding and using
-that data, how to proceed when some of that data is not available,
-and why we encourage everyone implementing simulations to have their
+We also compiled a series of guidelines for implementing
```
lauterbur: "compiled" -> "developed" or "formulated"? (Compiled implies they already existed in bits and pieces elsewhere.)

adding_species_manuscript.tex
```diff
@@ -852,159 +890,86 @@ \subsection*{\texorpdfstring{\emph{Bos taurus} (cattle)}{Bos taurus (cattle)}}
...
+population genomic simulations, which are based on
+insights from the community-driven process of expanding the \stdpopsim catalog.
+We list the basic requirements for generating a useful chromosome-level simulation for a given species, and explain the rationale behind these requirements.
+We discuss special considerations for collecting relevant information from the literature,
```
lauterbur: This is the third sentence in a row starting "We...". Perhaps "Furthermore, we..."?

lauterbur (commented):
Looks great, Ilan. My comments are mostly minor wording suggestions.

adding_species_manuscript.tex
igronau: I agree. It's better to focus on the main accomplishments here. I changed to:
> Through community-driven efforts, we expanded the number of species in the catalog by more than three-fold and broadened coverage across the tree of life.

adding_species_manuscript.tex
igronau: I also removed "another", since it's not clear what other learning modalities we're referring to. It also kind-of clashes with "also".
> Consequently, our intention with this paper is also to provide a learning modality to meet that need, ...

adding_species_manuscript.tex
igronau: Changed to:
> estimated by Keightley et al. (2009} for the closely related species, _Drosophila melanogaster_,
and ...

adding_species_manuscript.tex
igronau: formulated it is

adding_species_manuscript.tex
igronau: I tried to switch it up a bit like this:
> We also formulated a series of guidelines for implementing population genomic simulations, which are based on insights from the community-driven process of expanding the \stdpopsim catalog. **These guidelines specify** the basic requirements for generating a useful chromosome-level simulation for a given species, as well as the rationale behind these requirements. We **also** discuss special considerations for collecting relevant information from the literature, and what to do if some of that information is not available.

adding_species_manuscript.tex
igronau: I looked at the discussion again, but couldn't summarize it in a way which will highlight some useful lesson. I'll leave it as is for now, but maybe @petrelharp or @andrewkern will be able to add something here. I left the comment in the LaTex to remind us that this is left open

adding_species_manuscript.tex
igronau: missed that one. Will change to 
> . . . to conduct power analyses when designing empirical studies that use them.

adding_species_manuscript.tex
igronau: followed these suggestions, but kept the "so" (I like connecting this to the previous point directly)

adding_species_manuscript.tex
igronau: I don't see the harm in keeping it

adding_species_manuscript.tex
igronau: I think it's actually better to emphasized that we received feedback, so the long form seems better to me here

adding_species_manuscript.tex
igronau: changing this to "uncovered a vital need"

adding_species_manuscript.tex
igronau: This sentence is already quite long. We make this point in several places (the abstract and in the "updates" section), so I think we can stick with a brief statement here as is.

adding_species_manuscript.tex
igronau: I agree that your version is more accurate, but this is already a long statement, and I'd rather not add more here, unless it's necessary. I was trying to highlight two extremes of the spectrum: a one-off simulation, and a standardized stdpopsim simulation. Clearly, there is a scpectrum here, but the main point is made IMO by highlighting just the two extremes.

adding_species_manuscript.tex
igronau: I agree with your point. I went with this change:
> or with the intent of making the simulations available for future work by adding the appropriate models to \stdpopsim.

adding_species_manuscript.tex
igronau: very good point

PR #70, opened 2022-09-28
plr edits

petrelharp:
I made a pass through! Looks great - just minor edits. I'm happy to send this around.

### DIFF
--- a/adding_species_manuscript.tex
+++ b/adding_species_manuscript.tex
@@ -170,8 +170,8 @@ Here, we report on major updates made in the new release of \stdpopsim (version 
 which includes a significant expansion of the species catalog and substantial additions to simulation capabilities.
 Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
 Through community-driven efforts, we expanded the number of species in the catalog by more than three-fold and broadened coverage across the tree of life.
-In this paper, we aim to provide guidance for others aiming to do genome-scale simulations
-by reporting on the main lessons learned through this process.
+During the process of expanding the catalog,
+we have identified common sticking points and developed best practices for setting up genome-scale simulations. 
 We describe the input data required for generating a realistic simulation,
 suggest good practices for obtaining the relevant information,
 and discuss common pitfalls and major considerations.
@@ -229,7 +229,7 @@ encompassing population sizes and distributions, divergences, and gene flow, can
 dramatically affect patterns of genomic variation \citep{Teshima2006}. Thus
 species-specific estimates of these and other ecological and evolutionary parameters 
 (e.g., those governing the process of natural selection) 
-are important to realistic simulations.
+are important when generating realistic simulations.
 This presents challenges, especially to new researchers,
 as it takes a great deal of specialized knowledge not only to code the simulations themselves
 but also to find and choose appropriate estimates of the parameters underlying the simulation model.
@@ -311,7 +311,7 @@ If discrepancies are found in this process, they are resolved by discussion betw
 and if necessary with input of additional members of the community.
 This quality control process quite often finds subtle bugs \citep[e.g., as in][]{Ragsdale2020}
 or highlights parts of the model that are ambiguously defined by the literature sources.
-This increases the reproducibility of the resulting simulations in any downstream analysis.
+This increases the reliability and reproducibility of the resulting simulations in any downstream analysis.
 
 Another important goal of \stdpopsim is to promote whole-genome simulations,
 as opposed to the common practice of simulating many short segments  \citep[see, e.g.,][]{harris2016genetic}.
@@ -377,7 +377,7 @@ The hackathon was then structured based on feedback from these participants.
 One month before the hackathon, we organized a final workshop to prepare interested
 participants, by introducing them to  the process of developing
 a new species model and adding it to the \stdpopsim code base.
-Roughly 20 scientists participated in the hackathon (and are included as authors on this paper),
+Roughly 20 scientists participated in the hackathon (most of whom are included as authors on this paper),
 which resulted in the addition of 15 species to the \stdpopsim catalog
 (Figure \ref{fig:tree}).
 The catalog now includes
@@ -546,8 +546,8 @@ The development version of \stdpopsim enables simulation with selection
 by specifying genome annotations and distributions of fitness effects,
 as specified below.
 The ability to simulate chromosomes with realistic models of
-selection is still under development and will be finalized in the next release of \stdpopsim,
-and will be described in more detail in a future paper.
+selection is still under development. It will be finalized in the next release of \stdpopsim,
+and described in more detail in a future paper.
 
 \begin{enumerate}
 	\def\labelenumi{\arabic{enumi}.}
@@ -686,10 +686,10 @@ or study patterns of background selection,
 then it makes sense to simulate such pseudo-chromosomes.
 However, genetic correlations between different specific contigs lumped together in this way are obviously not accurate.
 So, if the main purpose of the simulation is to examine local patterns of genetic variation in loci of interest, then it may be more appropriate to simulate the relevant contigs separately (even if they are short), or to randomly sample several mappings of contigs to pseudo-chromosomes.
-% Finally, we note that for some purposes it may be sufficient to simulate a large number of unlinked sites \citep{Gutenkunst2009,Excoffier2013},
-% which can be generated without any sort of genome assembly.
-% However, this approach would not have the many benefits of whole-chromosome simulations,
-% which we promote in \stdpopsim.
+For some purposes it makes sense to simulate a large number of unlinked sites \citep{Gutenkunst2009,Excoffier2013},
+which can be generated without any sort of genome assembly.
+However, this approach would not have the benefits of whole-chromosome simulations.
+While some of the same considerations hold, a discussion of the advantages and disadvantages of such simulations goes beyond the scope of this paper.
 Ultimately, the recommended mode of simulation for species with partial genome builds depends on the intended use of the simulated genomes.
 
 \section*{Examples of added species}
@@ -763,7 +763,7 @@ All of these parameters were set in the appropriate source files in the \stdpops
 accompanied by the relevant citation information,
 and the model underwent the standard quality control process.
 %Ilan: any issues worth mentioning that were raised in the QC?
-The species may be refined in the future by adding more demographic models,
+The species entry may be refined in the future by adding more demographic models,
 updating or refining the recombination map,
 or updating the mutation rate estimates based on ones directly estimated for this species.
 Note that even if the mutation rate is ever updated,
@@ -853,9 +853,9 @@ we encourage wider adoption of ``code review'',
 i.e., for researchers implementing simulations to have their
 parameter choices and implementation reviewed by at least one other researcher.
 %% PLR: It's not totally clear which "these steps" are, and they are actually required for stdpopsim?
-% These steps can be followed if you are implementing a simulation independently for your own use,
-% or, as we encourage, when you add your simulation code to \stdpopsim,
-% to ensure its robustness and make it available for future standardized research.
+The guidelines in this paper can be followed when implementing a simulation independently for a single study,
+or (as we encourage others to do) when adding code to \stdpopsim,
+which helps to ensure its robustness and to make it available for future research.
 Currently, large-scale efforts such as the Earth Biogenome
 and its affiliated project networks are generating tens of thousands of genome
 assemblies. Each of these assemblies
--- a/adding_species_manuscript.tex
+++ b/adding_species_manuscript.tex
@@ -167,15 +167,11 @@ up-to-date information.
 The initial version of \stdpopsim focused on establishing this framework using 
 six well-characterized model species.
 Here, we report on major updates made in the new release of \stdpopsim (version 0.2),
-which includes a significant expansion of the species catalog and substantial additions to simulation capabilities that make it possible to simulate a broader range of species.
+which includes a significant expansion of the species catalog and substantial additions to simulation capabilities.
 Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
 Through community-driven efforts, we expanded the number of species in the catalog by more than three-fold and broadened coverage across the tree of life.
-Our experience throughout this process was that people are keen to put in the time and effort to include their study species,
-but that simple, clear guidance is vital.
-Consequently, our intention with this paper is also to provide a learning
-modality to meet that need,
-by reporting on the main lessons learned through this process
-for best practices in population genomic simulation.
+In this paper, we aim to provide guidance for others aiming to do genome-scale simulations
+by reporting on the main lessons learned through this process.
 We describe the input data required for generating a realistic simulation,
 suggest good practices for obtaining the relevant information,
 and discuss common pitfalls and major considerations.
@@ -226,14 +222,14 @@ specific and precise questions. Consequently, simulations must incorporate
 more and more detailed elements of a species' biology.
 Important elements include genomic features such as mutation and recombination
 rates that strongly affect genetic variation and haplotype structure
-\citep{Nachman2002}. These have particularly strong ramifications 
+\citep{Nachman2002}, which have particularly strong ramifications 
 when linked selection is important in the patterns of genomic diversity being studied \citep{Cutter2013}.
 Furthermore, the demographic history of a species,
 encompassing population sizes and distributions, divergences, and gene flow, can
 dramatically affect patterns of genomic variation \citep{Teshima2006}. Thus
 species-specific estimates of these and other ecological and evolutionary parameters 
 (e.g., those governing the process of natural selection) 
-are fundamentally important when developing simulations.
+are important to realistic simulations.
 This presents challenges, especially to new researchers,
 as it takes a great deal of specialized knowledge not only to code the simulations themselves
 but also to find and choose appropriate estimates of the parameters underlying the simulation model.
@@ -315,7 +311,7 @@ If discrepancies are found in this process, they are resolved by discussion betw
 and if necessary with input of additional members of the community.
 This quality control process quite often finds subtle bugs \citep[e.g., as in][]{Ragsdale2020}
 or highlights parts of the model that are ambiguously defined by the literature sources.
-This increases the reliability of the resulting simulations in any downstream analysis.
+This increases the reproducibility of the resulting simulations in any downstream analysis.
 
 Another important goal of \stdpopsim is to promote whole-genome simulations,
 as opposed to the common practice of simulating many short segments  \citep[see, e.g.,][]{harris2016genetic}.
@@ -341,11 +337,16 @@ effect on patterns of genome variation in myriad species
 \citep[e.g.,][]{McVicker2009,Charlesworth2012}. 
 In addition, the lengths of chromosome-scale shared haplotypes within and
 between populations provides valuable information on their demographic history.
-Demography inference methods that use such information,
+Demographic inference methods that use such information,
 such as MSMC \citep{Schiffels2020}, and IBDNe \citep{browning2015accurate},
 perform best on long genomic segments with realistic recombination rates.
 Chromosome-scale simulations are clearly required to test (or train) such methods,
 or to conduct power analyses when designing empirical studies that use them.
+
+\Stdpopsim makes these simulations available
+through only a single call to a command-line script
+or to a handful of lines of python code,
+along with citations to all primary sources underlying the simulation.
 
 
 \section*{Additions to \stdpopsim}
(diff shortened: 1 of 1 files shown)


adding_species_manuscript.tex
```diff
@@ -167,15 +167,11 @@ \section*{Abstract}
 The initial version of \stdpopsim focused on establishing this framework using 
 six well-characterized model species.
 Here, we report on major updates made in the new release of \stdpopsim (version 0.2),
-which includes a significant expansion of the species catalog and substantial additions to simulation capabilities that make it possible to simulate a broader range of species.
+which includes a significant expansion of the species catalog and substantial additions to simulation capabilities.
```
petrelharp, line 170: it's not yet clear why adding gene conversion makes it possible to simulate more species, so this was confusing, i thought

adding_species_manuscript.tex
```diff
@@ -167,15 +167,11 @@ \section*{Abstract}
...
 Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
 Through community-driven efforts, we expanded the number of species in the catalog by more than three-fold and broadened coverage across the tree of life.
-Our experience throughout this process was that people are keen to put in the time and effort to include their study species,
-but that simple, clear guidance is vital.
-Consequently, our intention with this paper is also to provide a learning
-modality to meet that need,
```
petrelharp, line 176: A "learning modality" is a *way* of learning things, whereas we have a set of guidelines. Also, the first sentence here seemed a bit condescending, so I rephrased.

adding_species_manuscript.tex
```diff
@@ -226,14 +222,14 @@ \section*{Introduction}
...
-\citep{Nachman2002}. These have particularly strong ramifications 
+\citep{Nachman2002}, which have particularly strong ramifications 
 when linked selection is important in the patterns of genomic diversity being studied \citep{Cutter2013}.
 Furthermore, the demographic history of a species,
 encompassing population sizes and distributions, divergences, and gene flow, can
 dramatically affect patterns of genomic variation \citep{Teshima2006}. Thus
 species-specific estimates of these and other ecological and evolutionary parameters 
 (e.g., those governing the process of natural selection) 
-are fundamentally important when developing simulations.
+are important to realistic simulations.
```
petrelharp: I don't think they're quite "fundamental" (i.e., everything else is built on them).

adding_species_manuscript.tex
```diff
@@ -341,12 +337,17 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
...
+Demographic inference methods that use such information,
 such as MSMC \citep{Schiffels2020}, and IBDNe \citep{browning2015accurate},
 perform best on long genomic segments with realistic recombination rates.
 Chromosome-scale simulations are clearly required to test (or train) such methods,
 or to conduct power analyses when designing empirical studies that use them.
 
+\Stdpopsim makes these simulations available
+through only a single call to a command-line script
+or to a handful of lines of python code,
+along with citations to all primary sources underlying the simulation.
```
petrelharp, line 349: It felt like we needed a conclusion sentence here.

adding_species_manuscript.tex
```diff
@@ -374,9 +375,9 @@ \section*{Additions to \stdpopsim}
 many of whom expressed interest in adding non-model species to \stdpopsim.
 The hackathon was then structured based on feedback from these participants.
 One month before the hackathon, we organized a final workshop to prepare interested
-participants for the hackathon, by introducing them to  the process of developing
+participants, by introducing them to  the process of developing
 a new species model and adding it to the \stdpopsim code base.
-Roughly 20 scientists participated in the hackathon,
+Roughly 20 scientists participated in the hackathon (and are included as authors on this paper),
```
petrelharp: I think they all are?

adding_species_manuscript.tex
```diff
@@ -500,7 +501,7 @@ \subsection*{Basic setup for chromosome-level simulations}
   At minimum, an average genome-wide recombination rate needs to be specified, which is typically available for well assembled genomes.
   For bacteria and archea, which primarily experience non-crossover recombination,
   the average tract length should also be specified
-  (see details in previous section).\\
+  (see details in previous section).
```
petrelharp, line 504: The indentation was weird with a linebreak, I thought - inline in the paragraph is okay.

adding_species_manuscript.tex
```diff
@@ -681,13 +683,13 @@ \subsection*{Filling out the missing pieces}
 %Ilan: tried to give a few concrete examples here
 If, for example, the main purpose of the simulation is to examine the distribution of lengths of shared haplotypes between individuals,
 or study patterns of background selection,
-then it makes sense to simulate such pseudo chromosomes.
-Note, however, that specific genetic correlations between different contigs lumped together can be highly inaccurate.
-So, if the main purpose of the simulation is to examine local patterns of genetic variation in loci of interest, then it would be more appropriate to simulate the relevant contigs separately (even if they are short), or to randomly sample several mappings of contigs to pseudo-chromosomes.
-Finally, we note that for some purposes it may be sufficient to simulate a large number of unlinked sites \citep{Gutenkunst2009,Excoffier2013},
```
petrelharp, line 687: Perhaps this is true, but tangential - I think to say this we should make clear what "some purposes" are, which is beyond the scope of this bit.

adding_species_manuscript.tex
```diff
@@ -732,8 +734,8 @@ \subsection*{\texorpdfstring{\emph{Anopheles gambiae} (mosquito)}{Anopheles gamb
 \citet{zheng1996integrated} (Figure \ref{fig:anogam}A).
 As direct estimates of \textbf{mutation rate} (e.g., via mutation accumulation) do not currently exist for \emph{Anopheles gambiae},
 we used the genome-wide average mutation rate of $\mu=3.5 \times 10^{-9}$ mutations per generation per site,
-estimated by \cite{Keightley2009} for the closely related species, \textit{Drosophila melanogaster},
-and used for analysis of \textit{A.~gambiae} data in \citet{Miles2017}.
+estimated by \cite{Keightley2009} for the fellow Dipteran \textit{Drosophila melanogaster},
```
petrelharp, line 737: mosquito-fly is what, like 40Mya?

petrelharp:
Here's a latexdiff.
[diff.pdf](https://github.com/popsim-consortium/adding-species-manuscript/files/9661648/diff.pdf)

adding_species_manuscript.tex
igronau: This is mostly a reference to bacterial recombination. Indeed, not all simulation capabilities that were added have to do with expanding the set of species. I'm flagging @lauterbur here, because I think it was text that she suggested.

adding_species_manuscript.tex
igronau: Agree, but "important to realistic simulations" reads a bit funny to me. How about "important when generating realistic simulations"?

adding_species_manuscript.tex
```diff
@@ -315,7 +311,7 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
 and if necessary with input of additional members of the community.
 This quality control process quite often finds subtle bugs \citep[e.g., as in][]{Ragsdale2020}
 or highlights parts of the model that are ambiguously defined by the literature sources.
-This increases the reliability of the resulting simulations in any downstream analysis.
+This increases the reproducibility of the resulting simulations in any downstream analysis.
```
igronau, line 315: I think both versions are fine, but you get reproducible simulations even if the code did not undergo any QC. IMO, the QC is mostly meant to improve the reliability of the simulations in the sense that the user can rely on the fact that they provide an honest implementation of the information in the cited sources.

adding_species_manuscript.tex
igronau: That's one for @lauterbur

adding_species_manuscript.tex
```diff
@@ -540,12 +541,13 @@ \subsection*{Basic setup for chromosome-level simulations}
...
-The current version of \stdpopsim enables simulation with selection
+The development version of \stdpopsim enables simulation with selection
 (using the \texttt{SLiM} engine)
 by specifying genome annotations and distributions of fitness effects,
 as specified below.
-We note that the ability to simulate chromosomes with realistic models of
-selection is still under development and will be finalized in the next release of \stdpopsim.
+The ability to simulate chromosomes with realistic models of
+selection is still under development and will be finalized in the next release of \stdpopsim,
+and will be described in more detail in a future paper.
```
igronau: you have two subsequent 'and's here. Should we switch the second one to a 'which' ("which will be described...")?

adding_species_manuscript.tex
igronau: The main reason to add this statement is to give people a last-resort option if they have popgen data but no genome assembly, like in RAD-SEQ studies. They might still be able to infer demographic histories and simulate data. I'm fine with leaving this out, so that this paragraph is focused on the more interesting cases of partial builds.

adding_species_manuscript.tex
igronau: 'closely related' is always very vague. We added this here just to signal that we chose D. mel because of its close relation to A. gam. "fellow Dipteran" is a better choice.

adding_species_manuscript.tex
```diff
@@ -761,7 +763,7 @@ \subsection*{\texorpdfstring{\emph{Anopheles gambiae} (mosquito)}{Anopheles gamb
 accompanied by the relevant citation information,
 and the model underwent the standard quality control process.
 %Ilan: any issues worth mentioning that were raised in the QC?
-The model may be refined in the future by adding more demographic models,
+The species may be refined in the future by adding more demographic models,
```
igronau: I don't assume that anyone is going to think we're considering modifying the actual mosquito species, but "refining a species" sounds a bit odd ;-).

adding_species_manuscript.tex
```diff
@@ -846,18 +849,20 @@ \section*{Conclusion}
...
-we encourage everyone implementing simulations to have their
-parameter choices and implementation reviewed by at least one other
-researcher.
-These steps can be followed if you are implementing a simulation independently for your own use,
-or, as we encourage, when you add your simulation code to \stdpopsim,
-to ensure its robustness and make it available for future standardized research.
+we encourage wider adoption of ``code review'',
+i.e., for researchers implementing simulations to have their
+parameter choices and implementation reviewed by at least one other researcher.
+%% PLR: It's not totally clear which "these steps" are, and they are actually required for stdpopsim?
```
igronau, line 855: "These steps" refers to our guidelines, and not just the QC part. This statement is important, but could be clarified by changing "These steps" to "The guidelines we provide in this paper". Some of these guidelines are strict requirements of stdpopsim and some are not. I think it's okay not to make this distinction here.

adding_species_manuscript.tex
igronau: I agree with changing "modality" to "guidelines", but I like preceding this with a statement that ties these guidelines to a need that came up during the expansion of the catalog. This gets a bit lost in the shorter version. I couldn't think of a good compromise. @lauterbur? Do you have a suggestion here?

igronau (commented):
Most edits look great. I left a few comments. I suggest that @lauterbur also takes a pass to resolve some of these remaining issues. They are all minor and appear to indicate oscillation of the editing process, which is a good sign that we're ready

lauterbur:
I skimmed these this morning and they looked good, I will take a closer
look within the next couple of hours.

adding_species_manuscript.tex
petrelharp: I understood it, and it's a good thought, but I didn't think it'd make sense to the reader at this point.

petrelharp: (happy to stick it back in though if someone thinks we should)

adding_species_manuscript.tex
```diff
@@ -167,15 +167,11 @@ \section*{Abstract}
...
 Features added to improve the realism of the simulated genomes include non-crossover recombination and provision of species-specific genomic annotations.
 Through community-driven efforts, we expanded the number of species in the catalog by more than three-fold and broadened coverage across the tree of life.
-Our experience throughout this process was that people are keen to put in the time and effort to include their study species,
-but that simple, clear guidance is vital.
-Consequently, our intention with this paper is also to provide a learning
-modality to meet that need,
-by reporting on the main lessons learned through this process
-for best practices in population genomic simulation.
+In this paper, we aim to provide guidance for others aiming to do genome-scale simulations
+by reporting on the main lessons learned through this process.
```
petrelharp:
```suggestion
During the process of expanding the catalog,
we have identified common sticking points and developed best practices for setting up genome-scale simulations. 
```

adding_species_manuscript.tex
petrelharp: See suggestion below.

adding_species_manuscript.tex
petrelharp:
```suggestion
are important when generating realistic simulations.
```

adding_species_manuscript.tex
petrelharp: Ah, okay. I was thinking about "reliable" as meaning that they'd, like, reliably run and not crash or something. Let's do "reliability and reproducibility". (since it is more reproducible if you know exactly what they're doing)

adding_species_manuscript.tex
```diff
@@ -315,7 +311,7 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
 and if necessary with input of additional members of the community.
 This quality control process quite often finds subtle bugs \citep[e.g., as in][]{Ragsdale2020}
 or highlights parts of the model that are ambiguously defined by the literature sources.
-This increases the reliability of the resulting simulations in any downstream analysis.
+This increases the reproducibility of the resulting simulations in any downstream analysis.
```
petrelharp:
```suggestion
This increases the reliability and reproducibility of the resulting simulations in any downstream analysis.
```

adding_species_manuscript.tex
petrelharp:
```suggestion
The ability to simulate chromosomes with realistic models of
selection is still under development. It will be finalized in the next release of \stdpopsim,
and described in more detail in a future paper.
```

adding_species_manuscript.tex
petrelharp: Good call, but I don't think we're saying people *can't* do single loci? We're not trying to review all possible simulation methods? (Or, if we do, we need some more context?)

adding_species_manuscript.tex
petrelharp: I changed it because "model" appeared too many times - how about "species entry"?

petrelharp:
```suggestion
The species entry may be refined in the future by adding more demographic models,
```

adding_species_manuscript.tex
```diff
@@ -846,18 +849,20 @@ \section*{Conclusion}
...
-These steps can be followed if you are implementing a simulation independently for your own use,
-or, as we encourage, when you add your simulation code to \stdpopsim,
-to ensure its robustness and make it available for future standardized research.
+we encourage wider adoption of ``code review'',
+i.e., for researchers implementing simulations to have their
+parameter choices and implementation reviewed by at least one other researcher.
+%% PLR: It's not totally clear which "these steps" are, and they are actually required for stdpopsim?
+% These steps can be followed if you are implementing a simulation independently for your own use,
+% or, as we encourage, when you add your simulation code to \stdpopsim,
+% to ensure its robustness and make it available for future standardized research.
```
petrelharp:
```suggestion
The guidelines in this paper can be followed when implementing a simulation independently for a single study,
or (as we encourage others to do) when adding code to \stdpopsim,
which helps to ensure its robustness and to make it available for future research.
```

adding_species_manuscript.tex
petrelharp: Ah, I see. I was shying a little away from saying "everyone please add all your simulation code to stdpopsim" since it can be hard to find QC'ers. But, how's this?

adding_species_manuscript.tex
lauterbur: Since we decided we're keeping the discussion of new features to those that are specific to the species catalog, my thought was to add something to that effect. (I think we say elsewhere that while there are also a bunch of additions for selection, we're explicitly not discussing those here, for instance.) I see your point that it isn't immediately clear with just this added clause. I don't know that it's confusing though, and is clarified somewhat in the next sentence.

adding_species_manuscript.tex
lauterbur: This works for me, so does the previous version.

adding_species_manuscript.tex
lauterbur: I do think they're fundamental for "realistic" simulations, but I'm fine with changing the language here.

adding_species_manuscript.tex
lauterbur: We have a few who declined authorship.

adding_species_manuscript.tex
igronau: "wide variety of other researchers and organisms"? Do we not consider all researchers to be organisms? ;-)
I'm not sure we need to emphasize the variety of researchers or species added here. Also, many of the "other researchers" are actually part of the author list, so we may not wish to "other" them. It might just be enough to say
> During the process of expanding the catalog, we have identified . . .  

The short sentence after that (In this paper, ...) becomes a bit redundant now. I think that the following sentence ("We describe ...") connects quite well this the previous one ("... setting up genome-wide simulations."). What do you say?

adding_species_manuscript.tex
igronau: Let's keep this commented out for now, and see if we miss this in future editing rounds. If needed, we can add this with some more detail, like saying that tasks such as demographic inference and DFE inference can be done using unlinked sites.

adding_species_manuscript.tex
igronau: Species entry sounds good

adding_species_manuscript.tex
lauterbur: I think the sentence is both useful and tangential - this paper isn't just about adding species to stdpopsim, it's also promoting standardized, species-specific simulations in general, which will appeal to anyone doing empirical research, including people with eg. RADSeq data. But those people aren't who the paper is for, and I think it's valid to make that pretty explicit. Perhaps "However, this approach would not have the many benefits of whole-chromosome simulations, which we promote in \stdpposim. While some of the same considerations hold, a discussion of such simulations goes beyond the scope of this paper."

adding_species_manuscript.tex
igronau: Looks good. I do think we should be encouraging people to add their code to stdpopsim. If they read this paper, then they know what our requirements and standards are, so I don't think we are risking open the door too wide here.

adding_species_manuscript.tex
igronau: The section in the paper does a good  in tying the added features to the expanded catalog. We don't necessarily have to explicitly make this point in the abstract. The problem with the statement that @petrelharp removed was that it implied that all (or most) of the features we added serve the purpose of allowing simulation of a wider range of species. In fact, this strictly true only for bacterial recombination. We also have finite site mutation models and gene conversion, which do not strictly allow modeling other species. So, I vote to keep @petrelharp's version here.

adding_species_manuscript.tex
igronau: I think that we can keep this statement as is even if not all participants agreed to be authors. I assume that everyone who ended contributing new species or demo models were included as authors.

petrelharp:
```suggestion
Roughly 20 scientists participated in the hackathon (most of whom are included as authors on this paper),
```

adding_species_manuscript.tex
igronau: I think that @petrelharp is raising another issue - that we should not suggest that simulation of independent sites is useful without explicitly saying when it is possibly useful. It's not like people will stop doing independent site simulations, and there are probably a few things that such simulations can benefit from by using our guidelines (QC, fitting a demographic model, mutation rate), but we don't necessarily need to go out of our way to make this point explicitly. My suggestion is to leave this part of the text commented out and return to it in a clean read. If it still looks to us like something's missing here, then we can figure a way to make this point the "right" way.

adding_species_manuscript.tex
```diff
@@ -681,13 +683,13 @@ \subsection*{Filling out the missing pieces}
...
-which can be generated without any sort of genome assembly.
-However, this approach would not have the many benefits of whole-chromosome simulations,
-which we promote in \stdpopsim.
+then it makes sense to simulate such pseudo-chromosomes.
+However, genetic correlations between different specific contigs lumped together in this way are obviously not accurate.
+So, if the main purpose of the simulation is to examine local patterns of genetic variation in loci of interest, then it may be more appropriate to simulate the relevant contigs separately (even if they are short), or to randomly sample several mappings of contigs to pseudo-chromosomes.
+% Finally, we note that for some purposes it may be sufficient to simulate a large number of unlinked sites \citep{Gutenkunst2009,Excoffier2013},
+% which can be generated without any sort of genome assembly.
+% However, this approach would not have the many benefits of whole-chromosome simulations,
+% which we promote in \stdpopsim.
```
petrelharp:
```suggestion
For some purposes it makes sense to simulate a large number of unlinked sites \citep{Gutenkunst2009,Excoffier2013},
which can be generated without any sort of genome assembly.
However, this approach would not have the benefits of whole-chromosome simulations.
While some of the same considerations hold, a discussion of the advantages and disadvantages of such simulations goes beyond the scope of this paper.
```

adding_species_manuscript.tex
petrelharp: See suggestion below.

adding_species_manuscript.tex
petrelharp: Good call; I've updated my suggestion above accordingly.

petrelharp:
OK, I've dealt with the comments, and I think this is ready to merge?

PR #72, opened 2022-09-29
Izabel's updates adding_species_manuscript.tex

izabelcavassim:
Adding some minor modifications to the manuscript. Very nice guys, I think it reads quite well. I felt like some of the portions were a bit repetitive, I tried to comment where I felt that way. But the information was given, I just think we need to be careful with the words we use so that it does not sound too informal. Happy to go through other rounds, and I am sorry for my absence.

### DIFF
--- a/adding_species_manuscript.tex
+++ b/adding_species_manuscript.tex
@@ -152,10 +152,10 @@
 \section*{Abstract}
 
 Simulation is a key tool in population genetics for both methods development and empirical research,
-but producing simulations that recapitulate even the main features of genomic datasets remains a major obstacle.
+but producing simulations that recapitulate the main features of genomic datasets remains a major obstacle.
 Today, more realistic simulations are possible thanks to large increases in
-the quantity of available data
-and the sophistication of inference and simulation software.
+genetic data,
+and to the sophistication of inference and simulation software.
 However, implementing these simulations still requires substantial time and specialized knowledge.
 These challenges are especially pronounced for simulating less well-studied species,
 since it is not always clear what level of realism is sufficient
@@ -195,7 +195,7 @@ providing the backbone for enormous increases in the amount of population-level 
 available for model and non-model species.
 These data are being used to answer questions across scales
 from deep evolutionary time to ongoing ecological dynamics.
-Methods that use these data, for example to infer demographic history and natural selection,
+Methods that use these data, e.g. to infer demographic history and natural selection,
 are also flourishing \citep{Beichman2018}.
 While past methods development focused on humans and a few key model systems such as \emph{Drosophila},
 more recent efforts are generalizing these methods to include 
@@ -209,23 +209,21 @@ and validating findings from empirical and theoretical research.
 For instance, simulations provide training data
 for inference methods based on machine learning \citep{Schrider2018} and
 Approximate Bayesian Computation \citep{Csillery2010}. They can also serve as
-baselines for further analyses: for example, simulations incorporating
+baselines for further analyses: simulations incorporating
 demographic history serve as null models when detecting selection \citep{Hsieh2016a}
 or seed downstream breeding program simulations \citep{Gaynor2020}.
-More recently, population genomic simulations have begun
-to be used to help guide conservation decisions for threatened species
+More recently, population genomic simulations have been used to help guide conservation decisions for threatened species
 \citep{Teixeira2021,kyriazis2022using}.
 
 Increasing amounts of data and sophistication of inference methods
-have enabled researchers to ask ever more
+have enabled researchers to ask even more
 specific and precise questions. Consequently, simulations must incorporate
 more and more detailed elements of a species' biology.
 Important elements include genomic features such as mutation and recombination
 rates that strongly affect genetic variation and haplotype structure
-\citep{Nachman2002}, which have particularly strong ramifications 
-when linked selection is important in the patterns of genomic diversity being studied \citep{Cutter2013}.
-Furthermore, the demographic history of a species,
-encompassing population sizes and distributions, divergences, and gene flow, can
+\citep{Nachman2002}. The inclusion of these genomic features is particularly important
+when linked selection is acting upon the patterns of genomic diversity being studied \citep{Cutter2013}.
+Furthermore, the demographic history of a species--encompassing population sizes and distributions, divergences, and gene flow--can
 dramatically affect patterns of genomic variation \citep{Teshima2006}. Thus
 species-specific estimates of these and other ecological and evolutionary parameters 
 (e.g., those governing the process of natural selection) 
@@ -314,7 +312,7 @@ or highlights parts of the model that are ambiguously defined by the literature 
 This increases the reliability and reproducibility of the resulting simulations in any downstream analysis.
 
 Another important goal of \stdpopsim is to promote whole-genome simulations,
-as opposed to the common practice of simulating many short segments  \citep[see, e.g.,][]{harris2016genetic}.
+as opposed to the common practice of simulating many short segments \citep[see, e.g.,][]{harris2016genetic}.
 Simulation of long sequences, on the order of $10^7$ bases,
 has until recently been computationally prohibitive,
 but this has changed with the development of modern simulation engines,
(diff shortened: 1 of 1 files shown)


adding_species_manuscript.tex
```diff
@@ -490,7 +488,7 @@ \subsection*{Basic setup for chromosome-level simulations}
   However, we anticipate future development will provide support for more complex, heterogeneous mutational processes,
   as these are easily specified in both the \texttt{SLiM} and \texttt{msprime} simulation engines.
   Such progress will further improve the realism of simulated genomes,
-  since mutation rates and processes are known to vary along the genome and through time \citep{Benzer1961,Ellegren2003,Supek2019}.
+  since mutation rates and evolutionary processes are known to vary along the genome and through time \citep{Benzer1961,Ellegren2003,Supek2019}.
```
igronau, line 492: The intent here is mutation processes and not just any processes. I seem to recall that @petrelharp added this, so I defer this to him.

adding_species_manuscript.tex
```diff
@@ -516,7 +514,7 @@ \subsection*{Basic setup for chromosome-level simulations}
   A given species might have more than one demographic model, fit from different data or by different methods.
   Thus, when selecting a demographic model, one should examine the data sources and methods used to obtain it to ensure that they are relevant to their study.
   At a minimum, simulation requires a single estimate of \textbf{effective population size}. This estimate, which may correspond to some sort of historical average effective population size,
-  should reproduce in simulation the average observed genetic diversity in that species. Note, however, that this average effective population size will not capture features of genetic variation that are caused by recent changes in population size and the presence of population structure \citep{MacLeod2013,Eldon2015}.
+  should translate in simulation to the average observed genetic diversity in that species. Note, however, that this average effective population size will not capture features of genetic variation that are caused by recent changes in population size and the presence of population structure \citep{MacLeod2013,Eldon2015}.
```
igronau: I prefer the original wording here, since 'translate' is a bit vague, whereas 'reproduce' means that the simulated data will have the same average genetic diversity as observed in real genomes.

adding_species_manuscript.tex
```diff
@@ -536,9 +534,9 @@ \subsection*{Basic setup for chromosome-level simulations}
 These five categories of parameters are sufficient for generating simulations
 under neutral evolution. Such simulations are useful for a number of purposes,
 but they cannot be used to model the influence of natural selection on patterns of genetic variation.
-As mentioned above, it is a widely appreciated fact that linked selection modulates
+As mentioned above, it is a widely appreciated the fact that linked selection modulates
```
igronau: Original wording seems better to me

adding_species_manuscript.tex
```diff
@@ -536,9 +534,9 @@ \subsection*{Basic setup for chromosome-level simulations}
...
 patterns of variation within genomes.
-Therefore, its incorporation into simulations is crucial for many purposes.
+Therefore, its incorporation into simulations is crucial for simulating realistic genomes.
```
igronau, line 539: I'm afraid that saying that it's crucial for realistic genomes will undercut the main message of this paper, which tries to set aside the issue of modeling selection. For this reason, 'many purposes' strikes a better balance, IMO.

adding_species_manuscript.tex
```diff
@@ -570,13 +568,13 @@ \subsection*{Basic setup for chromosome-level simulations}
 \subsection*{Extracting parameters from the literature}
 
 Simulations cannot of course precisely match reality, but in setting up simulations
-it is desireable to choose parameters that best reflect our current understanding.
+it is desireable to choose parameters that best reflect our current understanding of evolution.
```
igronau: 'understanding of evolution' is probably too big a goal here. Maybe replace with 'understanding of the evolutionary history of the species of interest'

adding_species_manuscript.tex
```diff
@@ -667,6 +665,7 @@ \subsection*{Filling out the missing pieces}
 So, the benefit of including species with very partial genome builds in \stdpopsim
 would be outweighed by the substantial extra burden on \stdpopsim maintainers as well as
 downstream users of these models.
+% Izabel: this bit above is a repetition, maybe delete or shrink? Or maybe when talking about it in the chromosome part just briefly mention it.
```
igronau: I like having a concluding sentence here. I think it does a decent job in summarizing the various points raised in this paragraph. maybe it could be shortened a bit...

igronau (commented):
Great edits, @izabelcavassim. Thanks. I left comments next to places that I think should be left with the original text or require a second set of eyes.

adding_species_manuscript.tex
izabelcavassim: ok so I think emphasizing that is mutation and not other process makes it more clear, because I thought that it was other processes as well. Maybe redundancy is good here?

adding_species_manuscript.tex
izabelcavassim: I just think it is too vague the many purposes, maybe including one example or two will address it better.

adding_species_manuscript.tex
izabelcavassim: yes, because I think it is missing what we understand

adding_species_manuscript.tex
izabelcavassim: But is it correct? I think it misses something, but I am not a native speaker :)

izabelcavassim: ah I see, 
```suggestion
As mentioned above, it is widely appreciated the fact that linked selection modulates
```

adding_species_manuscript.tex
igronau: I guess that 'mutation rates and mutation processes' makes sense.

adding_species_manuscript.tex
igronau: Still wrong IMO :-). 'It is a widely appreciated fact that X'
X is a fact, which is widely appreciated. Seems like a pretty straightforward phrasing. I suggest to keep it as it was, unless others also flag this one.

adding_species_manuscript.tex
igronau: Okay, I tried to think about specific purposes that would be short enough to describe, but it ended up running on too much of a tangent. I guess that your suggestion of 'simulating realistic genomes' strikes a good balance here. I read this now in context and think that there is little actual risk that this undercuts our message. So let's go with that.

adding_species_manuscript.tex
lauterbur: Another possibility that may feel less syntactically awkward, "mutation processes, including rates, are known to vary..."

adding_species_manuscript.tex
lauterbur: Either the original or the most recent suggestion are grammatically correct in my opinion (though I'd add a comma after "appreciated" in the most recent suggestion), but the original version is smoothest.

adding_species_manuscript.tex
```diff
@@ -152,10 +152,10 @@
 \section*{Abstract}
 
 Simulation is a key tool in population genetics for both methods development and empirical research,
-but producing simulations that recapitulate even the main features of genomic datasets remains a major obstacle.
+but producing simulations that recapitulate the main features of genomic datasets remains a major obstacle.
 Today, more realistic simulations are possible thanks to large increases in
-the quantity of available data
-and the sophistication of inference and simulation software.
+genetic data,
```
lauterbur: I like the increased specificity here, perhaps add "the quantity and quality of available data"

adding_species_manuscript.tex
```diff
@@ -209,23 +209,21 @@ \section*{Introduction}
 For instance, simulations provide training data
 for inference methods based on machine learning \citep{Schrider2018} and
 Approximate Bayesian Computation \citep{Csillery2010}. They can also serve as
-baselines for further analyses: for example, simulations incorporating
```
lauterbur, line 212: I would keep "for example," because there are potentially other ways they could be used as "baselines."

adding_species_manuscript.tex
```diff
@@ -209,23 +209,21 @@ \section*{Introduction}
...
+baselines for further analyses: simulations incorporating
 demographic history serve as null models when detecting selection \citep{Hsieh2016a}
 or seed downstream breeding program simulations \citep{Gaynor2020}.
-More recently, population genomic simulations have begun
-to be used to help guide conservation decisions for threatened species
+More recently, population genomic simulations have been used to help guide conservation decisions for threatened species
 \citep{Teixeira2021,kyriazis2022using}.
 
 Increasing amounts of data and sophistication of inference methods
-have enabled researchers to ask ever more
```
lauterbur: I prefer "ever," it implies that the specificity and precision will continue to increase in the future

adding_species_manuscript.tex
igronau: I suggest:
> the quantity and quality of available genetic data

adding_species_manuscript.tex
igronau: I like Elise's suggestion

adding_species_manuscript.tex
igronau: let's go with the original.

PR #76, opened 2022-10-03
Edit pass

grahamgower:


adding_species_manuscript.tex
```diff
@@ -430,7 +430,7 @@ \section*{Additions to \stdpopsim}
 
 Another important extension of \stdpopsim allows augmenting a genome assembly by genome annotations, such as coding regions, promoters, conserved elements, etc.
 These annotations can be used to simulate selection at a subset of sites (e.g., the annotated coding regions)
-using parametric distribution(s) of fitness effects.
+using a parametric distribution of fitness effects.
```
igronau: I think that the plural was used here on purpose to indicate that different annotated sets can be associated with different DFEs. Maybe we should use the plural and just remove the parentheses around the s?

igronau (commented):
Thanks, @grahamgower. I have only one comment about your suggested edit about the DFEs. Can you have another look there?

adding_species_manuscript.tex
```diff
@@ -205,8 +205,8 @@ \section*{Introduction}
 such as inbreeding or selfing \citep{Blischak2020}, skewed offspring
 distributions \citep{Montano2016}, and intense artificial selection \citep{MacLeod2013, MacLeod2014}.
 
-Simulations can be useful at all stages of this work --
-for planning studies, analyzing data, testing inference methods,
+Simulations can be useful at all stages of this work---for
```
lauterbur, line 208: Why three dashes here? Is this a latex thing?

adding_species_manuscript.tex
grahamgower: Sure, happy to go with the plural instead.

adding_species_manuscript.tex
grahamgower: An em-dash is three hyphens, `---`, while just two gives an en-dash. https://www.scribbr.com/language-rules/dashes/

lauterbur:
Gotcha, thanks!

PR #78, opened 2022-10-03
MFP suggestion to the manuscript

manolofperez:
Made a few suggestions to the text

### DIFF
--- a/adding_species_manuscript.tex
+++ b/adding_species_manuscript.tex
@@ -153,13 +153,13 @@
 
 \section*{Abstract}
 
-Simulation is a key tool in population genetics for both method development and empirical research,
+Simulation is a key tool in population genetics for both methods development and empirical research,
 but producing simulations that recapitulate the main features of genomic datasets remains a major obstacle.
 Today, more realistic simulations are possible thanks to large increases in
 the quantity and quality of available genetic data,
-and to the more sophisticated inference and simulation software.
+and to the sophistication of inference and simulation software.
 However, implementing these simulations still requires substantial time and specialized knowledge.
-These challenges are especially pronounced for simulating data for species that were not well-studied,
+These challenges are especially pronounced for simulating genomes for species that were not well-studied,
 since it is not always clear what information is required
 to produce simulations with a level of realism 
 to confidently answer a given question.
@@ -199,7 +199,7 @@ These data are being used to answer questions across scales
 from deep evolutionary time to ongoing ecological dynamics.
 Methods that use these data, e.g. to infer demographic history and natural selection,
 are also flourishing \citep{Beichman2018}.
-While past method development focused on humans and a few key model systems such as \emph{Drosophila},
+While past methods development focused on humans and a few key model systems such as \emph{Drosophila},
 more recent efforts are generalizing these methods to include 
 important population dynamics not initially accounted for,
 such as inbreeding or selfing \citep{Blischak2020}, skewed offspring
@@ -278,8 +278,8 @@ we highlight the ways in which the \stdpopsim framework eases the burden involve
 We begin by providing a brief overview of the importance of genome-wide simulations and the main rationale behind \stdpopsim;
 see \citet{Adrion2020} for more on the topic.
 The main objective of population genomic simulations is to recreate 
-patterns of sequence variation along the genome under known conditions
-that are modeled in a given species (or population) of interest.
+patterns of sequence variation along the genome under 
+the inferred evolutionary history of a given species.
 \Stdpopsim is built on top of the
 \texttt{msprime} \citep{Kelleher2016,Nelson2020,Baumdicker2022}
 and \texttt{SLiM} \citep{Haller2019} simulation engines,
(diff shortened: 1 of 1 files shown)


adding_species_manuscript.tex
```diff
@@ -153,18 +153,18 @@
 
 \section*{Abstract}
 
-Simulation is a key tool in population genetics for both methods development and empirical research,
+Simulation is a key tool in population genetics for both method development and empirical research,
 but producing simulations that recapitulate the main features of genomic datasets remains a major obstacle.
```
igronau: I think that methods development is actually the more common term to use in this context

adding_species_manuscript.tex
```diff
@@ -153,18 +153,18 @@
...
 Today, more realistic simulations are possible thanks to large increases in
 the quantity and quality of available genetic data,
-and to the sophistication of inference and simulation software.
+and to the more sophisticated inference and simulation software.
```
igronau: original text reads better to me here

adding_species_manuscript.tex
```diff
@@ -153,18 +153,18 @@
...
 However, implementing these simulations still requires substantial time and specialized knowledge.
-These challenges are especially pronounced for simulating less well-studied species,
-since it is not always clear what level of realism is sufficient
-to confidently answer a given question, or what information is required
-to produce simulations of that desired realism.
+These challenges are especially pronounced for simulating data for species that were not well-studied,
+since it is not always clear what information is required
+to produce simulations with a level of realism 
+to confidently answer a given question.
```
igronau, line 165: I like this edit. I would just use
> simulating genomes for species that were not well-studied

adding_species_manuscript.tex
```diff
@@ -199,11 +199,11 @@ \section*{Introduction}
 from deep evolutionary time to ongoing ecological dynamics.
 Methods that use these data, e.g. to infer demographic history and natural selection,
 are also flourishing \citep{Beichman2018}.
-While past methods development focused on humans and a few key model systems such as \emph{Drosophila},
+While past method development focused on humans and a few key model systems such as \emph{Drosophila},
 more recent efforts are generalizing these methods to include
```
igronau, line 203: also here, 'methods development' is the preferred term IMO

adding_species_manuscript.tex
```diff
@@ -390,7 +390,7 @@ \section*{Additions to \stdpopsim}
 two flowering plants including a crop (\textit{Helianthus annuus}),
 an algae (\textit{Chlamydomonas reinhardtii}),
 two bacteria,
-four primates and a common mammalian associate of primates (\textit{Canis familiaris}).
+four primates and a common mammalian associate of humans (\textit{Canis familiaris}).
```
igronau, line 393: I think that 'primates' was used here on purpose, as a little wink to the attentive reader

igronau (commented):
Thanks, @manolofperez. I like many of these edits, and wrote comments on others.

adding_species_manuscript.tex
lauterbur: Agreed. Although "primates" is less precise (dogs aren't exactly a common associate of chimps or lemurs, except as a potential predator), so someone without a sense of humor could potentially take issue. (Another thought - could this trip up someone whose native language is not English?)

adding_species_manuscript.tex
```diff
@@ -279,7 +279,7 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
 see \citet{Adrion2020} for more on the topic.
 The main objective of population genomic simulations is to recreate 
 patterns of sequence variation along the genome under known conditions
-that model a given species (or population) of interest.
+that are modeled in a given species (or population) of interest.
```
lauterbur: Perhaps "reflect a given species"?

adding_species_manuscript.tex
igronau: I guess we should avoid the inside joke here and switch to the more straightforward 'humans'

adding_species_manuscript.tex
```diff
@@ -333,7 +333,7 @@ \section*{The utility of \stdpopsim for genome-wide simulations}
 an even stronger effect. Selection acting on a small number of sites can
 indirectly influence levels and patterns of genetic variation at linked neutral sites,
 which has been shown to have a widespread
-effect on patterns of genome variation in myriad species
+effect on patterns of genome variation in a myriad of species
```
lauterbur, line 336: Either of these are grammatically correct, the first way ("in myriad species") is similar to saying "in many species," the second way ("in a myriad of species") is similar to "in a lot of species." Coin flip?

adding_species_manuscript.tex
igronau: All options sound equally good to me. I guess that if I have to say where the problem in this sentence is it is in the 'conditions'. What are these known conditions that we're referring to here? Maybe a more accurate description would be 'under the assumed evolutionary history of a given species'?

adding_species_manuscript.tex
igronau: I leave it up to you to decide....

adding_species_manuscript.tex
lauterbur: I think that's better, yes. The issue I see is that the sentence otherwise risks sounding redundant (even if it's not), as in "The main objective is to recreate patterns of sequence variation... that look like the variation in a given species." Your suggestion fixes that.

PR #81, opened 2022-10-06
Some author/funding edits + Dan's suggested edits

igronau:


adding_species_manuscript.tex
```diff
@@ -426,9 +429,10 @@ \section*{Additions to \stdpopsim}
 For example, the model for \emph{Drosophila melanogaster} has been updated in the \stdpopsim catalog to have a fraction of gene conversions of 0.83 (in all chromosomes with recombination) and an average tract length of 518 bases \citep{Comeron2012}.
 We note, however, that since non-crossover recombination incurs a high computation load in simulation,
 it is turned off by default in \stdpopsim, and must be explicitly invoked by the simulation model.
+Note that unless this option is explicitly turned on in simulation, the effective fraction of gene conversions is 0, even if a positive fraction is specified for that species.
```
igronau, line 434: I added this to clarify what happens if you specify a non-0 fraction of GC but do not explicitly turn on the GC flag. This is a bit technical, but potentially important. Pinging @petrelharp to check that I got it right.

petrelharp: Gee, isn't this what the previous sentence says? Maybe an even-more-clear and less redundant clarification than the sentence you wrote would be
```
We note, however, that since non-crossover recombination incurs a high computation load in simulation,
it is turned off by default in \stdpopsim, and must be explicitly invoked by the simulation model,
even for species like \textit{Drosohpila melanogaster} that have gene conversion parameters specified.
```

igronau: yes, this is better. Will change in next commit

igronau: @dschride - to answer your question, if we do not turn on the "GC flag" when simulating a species that has GC in its model (fraction of GC >0), the rate of recombination will be lower. We're just not simulating the GC events. I'd rather not get into these details in the manuscript. Just clarify that specifying a GC fraction > 0 doesn't automatically turn the GC flag on.

petrelharp: This is not right, @igronau - the GC flag does not affect the rate of crossing-over, and the rate of crossing-over is always equal to the specified recombination rate. I *think* this is what [the docs say](https://popsim-consortium.github.io/stdpopsim-docs/latest/api.html#gene-conversion-and-bacterial-recombination) but if someone has suggestions for how to make it more clear, please let us know.

petrelharp: Hm, unless you meant that the "effective rate of recombination" will  be lower in the same sense I wrote in the docs? There I said that
> enabling gene conversion effectively increases the rate of recombination, which can have a strong effect on runtimes
because gene conversion adds more locations where the trees change, like crossing-over does.

igronau: I might actually have gotten this wrong. I think I now understand this, and maybe this does require clarification in the text. My new suggestion is to clarify this using the Drosophila example. I pasted the text below, with my addition in bold
> For example, the model for Drosophila melanogaster has been updated in the stdpopsim catalog to have a fraction of gene conversions of 0.83 (in all chromosomes that undergo recombination) and an average tract length of 518 bases (Comeron et al., 2012). **This update does not affect the rate of crossover recombination, but it adds gene conversion events at a rate of 83:17 (roughly 5:1) relative to crossover recombination events.** We note that since non-crossover recombination incurs a high computation load in simulation, it is turned off by default in stdpopsim, and must be explicitly invoked by the simulation model. 

I think that this clarifies where the additional computation load comes from and also the fact that if you don't turn on the flag, you'll be modeling non-CO recombination at its correct rates, but you'll be ignoring the many GC events. What do you say, @dschride @petrelharp ?

igronau: See text in latest commit and let me know what you think

adding_species_manuscript.tex
```diff
@@ -471,26 +471,32 @@ \subsection*{Basic setup for chromosome-level simulations}
...
-  \citep{Amarasinghe2020,Amarasinghe2021} and concomitant advances in assembly pipelines
-  \citep{Chakraborty2016} are likely to boost these initiatives. 
-  When expanding the \stdpopsim catalog, we decided to focus on species with near-complete 
-  chromosome-level genome assemblies (i.e., close to one contig per chromosome).
-  This restriction was set mainly because species with less complete genome builds 
-  typically do not have genetic maps or good estimates of recombination rates, 
-  making chromosome-level simulation much less useful. 
-  Therefore, the utility of adding such species to the catalog does not justify the 
-  maintenance and storage burden incurred by the large number of contigs in these partial assemblies (see also discussion below).
+  When expanding the \stdpopsim catalog during the ``Growing the Zoo" hackathon,
```
igronau, line 475: Several people noted the redundancy between the discussion here and the one we have in the 'missing pieces' section,. So I decided to move all the discussion here, and make only a brief mention in the later section.

adding_species_manuscript.tex
```diff
@@ -540,24 +546,21 @@ \subsection*{Basic setup for chromosome-level simulations}
 These five categories of parameters are sufficient for generating simulations
 under neutral evolution. Such simulations are useful for a number of purposes,
 but they cannot be used to model the influence of natural selection on patterns of genetic variation.
```
igronau, line 549: removed this because it appears to be redundant with previous discussion. The main point here is how to model selection and not so much why it is important

adding_species_manuscript.tex
```diff
@@ -652,29 +656,13 @@ \subsection*{Filling out the missing pieces}
 \end{table}
 
 For many species it is difficult to obtain estimates of all necessary model parameters.
-Table \ref{tab:param-mod} provide suggestions for ways to deal with missing values of various central model parameters.
-The table also mentions the main discrepancies between the simulated data and real genomic data,
-which can be caused by misspecification of each parameter.
-
-Several researchers who participated in the ``Growing the Zoo'' hackathon wished to add species
```
igronau, line 655: merged the discussion in this paragraph into the 'chromosome-level genome assembly' paragraph in the prev section

adding_species_manuscript.tex
```diff
@@ -300,54 +301,52 @@ \section*{The utility of \stdpopsim for chromosome-scale simulations}
...
-such as MSMC \citep{Schiffels2020}, and IBDNe \citep{browning2015accurate},
+such as MSMC \citep{Schiffels2020} and IBDNe \citep{browning2015accurate},
 perform best on long genomic segments with realistic recombination rates.
 Chromosome-scale simulations are clearly required to test (or train) such methods,
 or to conduct power analyses when designing empirical studies that use them.
 
-\Stdpopsim makes these simulations available
-through only a single call to a command-line script
-or to a handful of lines of python code,
+With \stdpopsim, such simulations are available with just a single call to a command-line script or with execution of handful of lines of Python code,
```
bhaller: "a" handful, add "a"

adding_species_manuscript.tex
```diff
@@ -426,42 +425,43 @@ \section*{Additions to \stdpopsim}
...
+For example, the model for \emph{Drosophila melanogaster} has been updated in the \stdpopsim catalog to have a fraction of gene conversions of 0.83 (in all chromosomes that undergo recombination) and an average tract length of 518 bases \citep{Comeron2012}.
+This update does not affect the rate of crossover recombination, but it adds gene conversion events at a ratio of 83:17 (roughly 5:1) relative to crossover recombination events.
+We note that since non-crossover recombination incurs a high computation load in simulation,
+it is turned off by default in \stdpopsim, 
+and must be explicitly invoked by the simulation model.
 
 
-Another important extension of \stdpopsim allows augmenting a genome assembly with genome annotations, such as coding regions, promoters, conserved elements, etc.
-These annotations can be used to simulate selection at a subset of sites (e.g., the annotated coding regions)
+Another important extension of \stdpopsim allows augmenting a genome assembly with genome annotations, such as coding regions, promoters, conserved elements.
```
bhaller: "and" conserved, add "and"

adding_species_manuscript.tex
```diff
@@ -502,34 +508,34 @@ \subsection*{Basic setup for chromosome-level simulations}
...
   \textbf{A demographic model} describing 
-  ancestral population sizes, split times and migration rates.
+  ancestral population sizes, split times, and migration rates.
   Selection of a reasonable demographic model is often crucial,
   since misspecification of the model can generate unrealistic patterns of genetic variation that will affect downstream analyses \citep[e.g.,][]{Navascues2009}.
   % PLR: an odd citation for that?
   % ILAN: I tried looking for a better one. Will try to crowdsource this
   A given species might have more than one demographic model, fit from different data or by different methods.
-  Thus, when selecting a demographic model, one should examine the data sources and methods used to obtain it to ensure that they are relevant to their study.
+  Thus, when selecting a demographic model, one should examine the data sources and methods used to obtain it to ensure that they are relevant to the study at hand.
```
bhaller, line 528: good solution, I will remember this!

adding_species_manuscript.tex
```diff
@@ -652,29 +656,13 @@ \subsection*{Filling out the missing pieces}
...
-Finally, each species requires validation of its code before it is added to the \stdpopsim catalog,
-as well as long-term maintenance to keep it up-to-date with changes made to the \stdpopsim framework.
-So, the benefit of including species with very partial genome builds in \stdpopsim
-would be outweighed by the substantial extra burden on \stdpopsim maintainers as well as
-downstream users of these models.
-
-
-%Ilan: applied some important edits in this paragraph to better address some of the delicate situations and options we suggest for them 
-That being said, simulation is still potentially useful for species with partial genome builds.
+Table \ref{tab:param-mod} provides suggestions for ways to deal with missing values of various central model parameters.
```
bhaller: maybe delete "central"?  not really clear what is meant...?

adding_species_manuscript.tex
```diff
@@ -652,29 +656,13 @@ \subsection*{Filling out the missing pieces}
...
+The table also mentions possible consequences of misspecification of each parameter.
+
+
+There are additional considerations that apply to species with partial genome builds.
+We discussed above our choice in \stdpopsim to focus on species with chromosome-level assemblies.
+However, simulation is still potentially useful for species with partial genome builds.
```
bhaller: How about deleting the first sentence, "There are additional considerations that apply to species with partial genome builds.", and adding to the end of this sentence ", with some additional considerations".  (note the comma at the start of the added text)

adding_species_manuscript.tex
dschride: Yes I think that is perfectly clear.

adding_species_manuscript.tex
igronau: I can do that. The main reason I added the first sentence was to transition from the previous short paragraph on considerations when some parameters are missing. Don't you think that we need a sentence to connect "missing parameters" with "partial assemblies"?

bhaller: Hmm, feels like it would be OK to me; since it's a new paragraph, a shift in topic is expected.  But any edit is fine, as long as it gets rid of the large amount of redundancy between the first and third sentences above.  Maybe you can consolidate them into the first sentence, instead of consolidating them into the third sentence as I suggested.

adding_species_manuscript.tex
```diff
@@ -660,9 +660,8 @@ \subsection*{Filling in the missing pieces}
 The table also mentions possible consequences of misspecification of each parameter.
 
 
-There are additional considerations that apply to species with partial genome builds.
-We discussed above our choice in \stdpopsim to focus on species with chromosome-level assemblies.
-However, simulation is still potentially useful for species with partial genome builds.
+In some cases, one may wish to generate simulations for a species with a partial genome build.
+Despite the focus of \stdpopsim  on species with chromosome-level assemblies (see discussion above), simulation is still potentially useful for species with less complete assemblies, with some important considerations to keep in mind.
 Longer contigs or scaffolds in these builds can be simulated separately and independently.
```
igronau, line 666: @bhaller What do you think of this? The 2nd sentence is a bot long, but I think it works.

adding_species_manuscript.tex
igronau: I made another edit trying to fix this. See below

adding_species_manuscript.tex
bhaller: Sure, works for me.  :->

adding_species_manuscript.tex
lauterbur: Is it necessary to clarify that 83:17 is "roughly 5:1"?

lauterbur (commented):
LGTM

adding_species_manuscript.tex
igronau: I removed this in my latest commit

PR #89, opened 2022-10-19
Fig1 caption tweak

igronau:


igronau:
This PR ended up being a bit messy and can be removed without being merged. See PR #90 , which is cleaner

PR #95, opened 2023-03-07
Reference swap

lauterbur:
Got rid of Leigh et al. citation and replaced with Hsieh et al. citation since Leigh only mentions stdpopsim rather than using it


petrelharp:
Good catch! (I'd merge but am not sure if you have other plans...)

